PR #80, opened 2015-04-26
API improvements

gentlecat:
/cc @mayhem, @ianmcorvidae

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -18,6 +18,8 @@ of Linux.**
 
 6. Standard Development Tools™ (``sudo apt-get install build-essential``)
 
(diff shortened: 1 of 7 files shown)


gentlecat:
Maybe we should remove ``replication/`` parts from endpoint paths and just leave it like that:

* ``/api/musicbrainz/replication-<PACKET_NUMBER>.tar.bz2``
* ``/api/musicbrainz/replication-daily-<PACKET_NUMBER>.tar.bz2``
* ``/api/musicbrainz/replication-weekly-<PACKET_NUMBER>.tar.bz2``

What do you think?

mayhem:
:+1:

PR #81, opened 2015-04-27
Use Flask-Login to manage admin access

gentlecat:
I replaced Basic HTTP Auth for admin pages with Flask-Login. It is now using MusicBrainz OAuth to check who is logged in and determines whether admin pages are accessible by checking list of allowed users in the config file. :customs:

This should make it easier to work with. No need for username/password combinations.


gentlecat:
I switched to that branch on rika. See http://metabrainz.mbsandbox.org/.

metabrainz/config.py.example
```diff
@@ -48,6 +48,14 @@ MUSICBRAINZ_CLIENT_ID = ""
 MUSICBRAINZ_CLIENT_SECRET = ""
 
 
+# ADMIN SECTION
+# This is a list of MusicBrainz usernames of people that are allowed to access
+# admin section of the website. Usernames are case-sensetive!
+ADMINS = [
+    'rob',
+]
```
mayhem: Please add yourself as well. It will be nice to have two people be able to do things.

mayhem:
:+1: I really like this one -- one less password for me to forget!

PR #83, opened 2015-04-28
Return shorter responses from the API when errors occur

gentlecat:
I think there's no need to transfer all our fancy HTML on default error pages to API users.


mayhem:
:+1:

PR #85, opened 2015-04-29
Fixed problem with incorrect scheme in external URLs

gentlecat:


mayhem:
Gah, that looked tedious. :+1:

Issue #94, opened 2015-05-04
Sign up vs Login


mayhem:
Right now if the user wants to edit their account, they must sign in via the "sign up" link. Which is confusing if you have already signed up.

Maybe: At the bottom we have a Sign In link and in the navbar we have the "Support Us" link? (#90)

PR #95, opened 2015-05-04
Updated parts of the interface related to user accounts

### LINKED ISSUE #90: Sign Up -> Support us!
We should use say "Support us" instead of "Sign up". Please change the text and then add a "Support us" navbar item in a contrasting color to the navbar all the way on the right. Thus button should only be shown when there isn't a user logged in. Maybe we can show the login name in that spot when a user is logged in?

gentlecat:
*Fixes #90, #94.*

### DIFF
--- a/metabrainz/static/css/main.less
+++ b/metabrainz/static/css/main.less
@@ -38,6 +38,12 @@ body {
 img { .img-responsive(); }
 img.pull-right { padding-left: 15px; }
 
+
+a.support-link {
+  color: lighten(@green, 30%) !important;
+}
+
+
 // Home page
 #homepage {
   @section-margin: 40px;
--- a/metabrainz/static/css/main.less
+++ b/metabrainz/static/css/main.less
@@ -39,8 +39,13 @@ img { .img-responsive(); }
 img.pull-right { padding-left: 15px; }
 
 
-a.support-link {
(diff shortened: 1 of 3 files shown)


metabrainz/static/css/main.less
```diff
@@ -38,6 +38,12 @@ body {
 img { .img-responsive(); }
 img.pull-right { padding-left: 15px; }
 
+
+a.support-link {
+  color: lighten(@green, 30%) !important;
+}
```
mayhem: Let's please try a red background (same as the under construction red bar) with a bold white text.

metabrainz/templates/navbar.html
```diff
@@ -29,6 +29,25 @@
         <li><a href="{{ url_for('donations.index') }}">Donate</a></li>
         <li><a href="http://blog.musicbrainz.org/category/metabrainz/">Blog</a></li>
       </ul>
+
+      {# Not showing user stuff on error pages. If attempt to load user info fails there, page will not render. #}
+      {% if not error %}
+        <ul class="nav navbar-nav navbar-right">
+          {% if not current_user or current_user.is_anonymous() %}
+            <li><a class="support-link" href="{{ url_for('users.signup') }}">Support us</a></li>
```
mayhem: This should link to /supporters/tiers

PR #96, opened 2015-05-05
Added 'in_deadbeat_club' flag into User model

gentlecat:
```sql
ALTER TABLE "user" ADD COLUMN in_deadbeat_club BOOLEAN;
UPDATE "user" SET in_deadbeat_club = FALSE;
ALTER TABLE "user" ALTER COLUMN in_deadbeat_club SET NOT NULL;
```


mayhem:
:+1:

PR #106, opened 2015-05-06
Profile improvements

mayhem:
Add a section that gives the user permission or not to use our data (it depends on their account standing). 

Add more details to their profile. The display could be cleaned up a bit.

Also, somewhat OT, allow three lines of description in the tiers overview.

### DIFF
--- a/metabrainz/templates/users/profile.html
+++ b/metabrainz/templates/users/profile.html
@@ -18,20 +18,9 @@
     </span>
   </p>
 
-  <h3>Contact information</h3>
-  <p>
-    <strong>Name:</strong> {{ current_user.contact_name }}<br />
-    <strong>Email:</strong> {{ current_user.contact_email }}<br />
-  </p>
-  <p>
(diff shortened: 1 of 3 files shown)


metabrainz/templates/users/profile.html
```diff
@@ -18,10 +18,50 @@
     </span>
   </p>
 
+  {% if current_user.is_commercial %}
+      {% if current_user.state == 'active' and current_user.good_standing %}
```
gentlecat: What about `in_deadbeat_club`?

mayhem: That will never really be seen by outsiders. Someone is in the deadbeat club because we added them to it. When they decide to start supporting us, I'll move them to the correct tier before I change the user.

gentlecat: OK.

gentlecat:
Looks good. :shipit:

PR #114, opened 2015-05-07
Added amount_pledged column

### LINKED ISSUE #105: Add database field to track actual amount pledged
Right now we track the tiers, but those only define thresholds. In the signup page we should include a field (amount_pledged) that allows the user to pick an actual amount to support us with. 

The text for this should read:

"If you would like to support us with more than $tier_threshold_amount, please enter the actual amount here:"

This field should initially be populated with $tier_threshold_amount, and this field should not be user editable.

gentlecat:
```sql
ALTER TABLE "user" ADD COLUMN amount_pledged NUMERIC(11,2);
```

*Fixes #105.*


mayhem:
:+1:

PR #117, opened 2015-05-09
Using IP addresses for API abuse detection

### LINKED ISSUE #99: Hourly access threshold exceeded
I received this message when I tried to have a replicated server catch up, meaning that it would need to copy a pile of replication packets in one hour. :(

We should tweak this check to only trigger if packets are being downloaded from a number of different IPs. For instance, if we have currently have more than 10 downloads for a token trigger this warning, we should change it to trigger it if more than 10 different IP addresses download it in an hour.

gentlecat:
Schema update:
```sql
ALTER TABLE access_log ADD COLUMN ip_address INET;
```

*Fixes #99.*

### DIFF
--- a/metabrainz/model/access_log.py
+++ b/metabrainz/model/access_log.py
@@ -1,11 +1,14 @@
 from metabrainz.model import db
 from metabrainz.mail import send_mail
 from metabrainz import cache
+from sqlalchemy.dialects import postgres
 from datetime import datetime, timedelta
 from flask import current_app
 import logging
+import pytz
 
-HOURLY_ALERT_THRESHOLD = 10  # Max number of requests that can be done before admins get an alert
+CLEANUP_RANGE_MINUTES = 60
+DIFFERENT_IP_LIMIT = 10
 
 
 class AccessLog(db.Model):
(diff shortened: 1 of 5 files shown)


metabrainz/model/access_log.py
```diff
@@ -1,11 +1,14 @@
...
 from metabrainz import cache
+from sqlalchemy.dialects import postgres
 from datetime import datetime, timedelta
 from flask import current_app
 import logging
+import pytz
 
-HOURLY_ALERT_THRESHOLD = 10  # Max number of requests that can be done before admins get an alert
+CLEANUP_RANGE_MINUTES = 60
+DIFFERENT_IP_LIMIT = 10
```
gentlecat: Can adjust these two variables if they should  be different.

metabrainz/model/access_log.py
```diff
@@ -13,40 +16,53 @@ class AccessLog(db.Model):
...
-                                 cls.token == access_token).count()
-        if count > HOURLY_ALERT_THRESHOLD:
-            msg = "Hourly access threshold exceeded for token %s" % access_token
+        count = cls.query \
+            .distinct(cls.ip_address) \
+            .filter(cls.timestamp > datetime.now(pytz.utc) - timedelta(minutes=CLEANUP_RANGE_MINUTES),
+                    cls.token == access_token) \
+            .count()
+        if count > DIFFERENT_IP_LIMIT:
+            msg = ("Hourly access threshold exceeded for token %s\n\n"
```
mayhem: I would love the see the count in the message. Exactly how many times has it been used? (might save me a trip to the web page to check)

mayhem:
Aside from the request about count in the email, this looks good. please add count, then merge. Also, is there a cron job that needs to be setup for the IP addr cleanup?

gentlecat:
Cleanup is done when new access records are added.

mayhem:
Oh and :+1: :mega:

PR #127, opened 2015-05-12
Fixed PayPal IPN endpoint

gentlecat:
IPN endpoint should work better now. Though I still can't get my payments to go though in sandbox. Every single one is `pending`. `pending_reason` is set to `unilateral` which, [according to docs](https://developer.paypal.com/webapps/developer/docs/classic/ipn/integration-guide/IPNandPDTVariables/#id091EB04C0HS), means that payment was made to an email address that is not yet registered or confirmed. I can't find a solution to this.


mayhem:
:+1:

PR #132, opened 2015-05-12
Add AcoustID logo

lalinsky:


mayhem:
Thanks!

gentlecat:
@mayhem, maybe we should keep logos that are referenced from the DB on a separate server that is easier to update? It seems like a bad idea to me to update repository every time there is a new user or if old one needs an update.

PR #152, opened 2015-05-17
Grouping donations from the same person in get_biggest_donations method

### LINKED ISSUE #150: Listing donations by amount should sum(amount)
The old metabrainz site uses this query to show the top donors by amount:

https://github.com/metabrainz/metabrainz-server/blob/master/lib/MetaBrainz/Data/Donation.pm#L105

(Note that there are no duplicates in the data -- each of those donations has a different date)

gentlecat:
*Fixes #150.*


mayhem:
:+1:

I'm glad you understand sqlalchemy. :-)

gentlecat:
I don't. :)

Issue #155, opened 2015-05-18
Deadlock when trying to write IP address to log


mayhem:
On the flipside, logging is working. :)

As for fixing this error, if it occurs again, the error should be logged, but not be fatal. Let the request proceed successfully.

```
   Message type: ERROR
   Location: /home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/flask/app.py:1423
   Module: app
   Function: log_exception
   Time: 2015-05-18 04:10:04,323
   Message:
   Exception on /api/musicbrainz/replication-86400.tar.bz2 [GET]

Traceback (most recent call last):
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
   response = self.full_dispatch_request()
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
   rv = self.handle_user_exception(e)
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
   reraise(exc_type, exc_value, tb)
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
   rv = self.dispatch_request()
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
   return self.view_functions[rule.endpoint](**req.view_args)
 File "./metabrainz/api/decorators.py", line 16, in decorated
   return f(*args, **kwargs)
 File "./metabrainz/api/decorators.py", line 30, in decorated
   AccessLog.create_record(request.args.get('token'), ip_addr)
 File "./metabrainz/model/access_log.py", line 49, in create_record
   cls.remove_old_ip_addr_records()
 File "./metabrainz/model/access_log.py", line 80, in remove_old_ip_addr_records
   update({cls.ip_address: None})
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2949, in update
   update_op.exec_()
 File "/home/metabrainz/metabrainz.org/venv/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 1074, in exec_
   self._do_exec()
[...]

PR #156, opened 2015-05-18
Disabled operation that causes deadlock

gentlecat:
See #155.


gentlecat:
Feel free to merge this if it gets bad. I'll take a look at ways to fix this when the dust settles.

mayhem:
:+1:

PR #164, opened 2015-05-18
Replace circle Tomahawk logo with textual logo

navap:
https://metabrainz.org/supporters/tiers/8 looks a little messed up with the giant circle Tomahawk logo. I thought this one would fit better: https://www.tomahawk-player.org/assets/images/pages/about/tomahawk-logo.png.

I notice the /img/logos/supporters directory isn't in this repo (yet?), is that an oversight? Hopefully this patch works.


gentlecat:
I think we have a separate repo for these logos.

mayhem:
Thanks for the link -- but the image needs to go into a different private repo. I've added it there. Thanks!

PR #167, opened 2015-05-18
Include external links in the carousel

navap:
Note that the AcousticBrainz site is not accessible over TLS.


gentlecat:
LGTM :shipit:

PR #168, opened 2015-05-19
Fix wording

navap:


metabrainz/templates/api/info.html
```diff
@@ -48,11 +48,11 @@
       </code>
     </p>
     <p>
-      It is possible to get signature for each replication packet. Just replace
+      It is possible to get a signature for each replication packet. Just replace
       <code>.tar.bz2</code> with <code>.asc</code>.
     </p>
     <p>
-      You can get info about latest replication packets from this endpoint:<br />
+      You can get find the latest replication packet numbers from this endpoint:<br />
```
gentlecat: > get find

I guess one of these is unnecessary. :)

mayhem:
I fixed your duplicate word and merged this. Thanks!

metabrainz/templates/api/info.html
navap: Whoops, good catch!

PR #170, opened 2015-05-22
Replication check

mayhem:
This PR creates a nagios style endpoint for checking the completeness of validity of the current set of replication packets. Nagios wants to have a single line of output that starts with OK, WARNING, CRITICAL or UNKNOWN; the rest of the line is for humans to consume and should be telling of that the problem is.

This PR checks to make sure that all files are there and that the latest file is less then 4 hours old. We may be able to tighten this check, but lets leave it there for now.

### DIFF
--- a/metabrainz/api/views.py
+++ b/metabrainz/api/views.py
@@ -6,6 +6,7 @@ from metabrainz.api.decorators import token_required, tracked
 import logging
 import re
 import os
+import time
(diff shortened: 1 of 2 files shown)


gentlecat:
Looks good apart from some PEP 8 things. I can push a fix for that if you want.
:ship:

PR #173, opened 2015-06-07
Improved user lists

gentlecat:


gentlecat:
*http://metabrainz.mbsandbox.org/supporters*

metabrainz/templates/users/supporters-list.html
```diff
@@ -24,7 +24,7 @@ <h1 class="page-title">Supporters</h1>
         <div class="users-grid">
           {% for user in featured_users %}
             <div class="user">
-              <a href="{{ url_for('users.tier', tier_id=tier.id) }}">
+              {% if user.website_url %}<a href="{{ user.website_url }}">{% endif %}
```
gentlecat: We are currently not showing there which users have bad standing. Not sure if we should do it there like on tier page (see my change to *metabrainz/templates/users/tier.html* in this pull request).

mayhem: I think we should. I doubt many people will click through to the individual tiers page. I think I might toss Zvooq into this bin, so we can see how it looks on the site. Please add this and then feel free to merge.

mayhem:
:+1:

PR #177, opened 2015-07-14
MEB-5: Added link to user model editor into admin/users/index.html template

gentlecat:
*Fixes [MEB-5](http://tickets.musicbrainz.org/browse/MEB-5).*


mayhem:
:+1:

PR #180, opened 2015-07-14
MEB-11: Removed payment_method column from User model

gentlecat:
```sql
ALTER TABLE "user" DROP COLUMN payment_method;
```

*Fixes [MEB-11](http://tickets.musicbrainz.org/browse/MEB-11).*


mayhem:
Doesn't removing code feel good? :+1:

PR #185, opened 2015-07-26
Added autofocus attribute to some inputs in admin section

gentlecat:
*Fixes [MEB-32](http://tickets.musicbrainz.org/browse/MEB-32).* :rainbow:


mayhem:
:+1:

PR #195, opened 2015-09-28
Check if user already exists before trying to create a duplicate account

gentlecat:
This should fix http://sentry.roman.zone/sentry/meb-production/group/70/.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 Flask == 0.10.1
 Flask-Testing == 0.4.2
-Flask-WTF == 0.11
+Flask-WTF == 0.12
 Flask-Login == 0.2.11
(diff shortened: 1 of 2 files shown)


mayhem:
:+1: 

That was a captivating PR -- about as captivating as the eclipse has been. :-p

mayhem:
Err, sorry :sheep:

requirements.txt
```diff
@@ -1,6 +1,6 @@
 Flask == 0.10.1
 Flask-Testing == 0.4.2
-Flask-WTF == 0.11
+Flask-WTF == 0.12
```
gentlecat: This should fix http://sentry.roman.zone/sentry/meb-production/group/69/.

PR #208, opened 2016-01-24
MEB-45: Projects page

caroline-gschwend:


metabrainz/templates/index/projects.html
```diff
@@ -0,0 +1,134 @@
...
+<div id="projects">
+  
+  <div class="media">
+    <div class="media-left media-middle">
+      <a href="https://musicbrainz.org/">
+        <img class="media-object" src="{{ url_for('static', filename='img/projects/MBnt.png') }}" alt="MusicBrainz logo">
+      </a>
+    </div>
+    <div class="media-body">
+      <h4 class="media-heading"><strong>MusicBrainz</strong></h4>
```
Freso: I feel like it's superfluous to have `<strong>` within an `<h4>`. Just mark `h4`s as `font-weight: bold;` or whatever it is. Possibly `h4.media-heading`. Ask Roman or Leo.

metabrainz/static/css/main.less
```diff
@@ -211,6 +211,10 @@ h1.page-title, h2.page-title {
   }
 }
 
+// Projects page
+#projects {
+  img { width: 90px; }
```
Freso: Are we sure that we want absolutely all images on this page to be this wide? I think it'd probably be better to add a `logo` class or something like that, and use that class for the images. (And apply this `width` to the class instead of the `` element in general.)

metabrainz/static/css/theme/boostrap/boostrap.less
```diff
@@ -35,7 +35,7 @@
 @import "thumbnails.less";
 @import "alerts.less";
 //@import "progress-bars.less";
-//@import "media.less";
+@import "media.less";
```
Freso: I'm not saying this is wrong, I'm just curious about why you did this. :)

LeoVerto: \o That was my decision.
This is exactly the example use case for Bootstrap [media objects](https://getbootstrap.com/components/#media) and the previously used grid model styling ran into some issues.

When this site was initially designed someone commented that line out thus disabling the relevant LESS because it simply didn't use media objects at that time.

metabrainz/templates/index/projects.html
```diff
@@ -0,0 +1,134 @@
...
+{% block title %}Projects - MetaBrainz Foundation{% endblock %}
+
+{% block content %}
+<h1 class="page-title">The MetaBrainz Foundation Projects</h1>
+<div id="projects">
+  
+  <div class="media">
+    <div class="media-left media-middle">
+      <a href="https://musicbrainz.org/">
+        <img class="media-object logo" src="{{ url_for('static', filename='img/projects/MBnt.png') }}" alt="MusicBrainz logo">
```
Freso: What is the "nt" in the filenames? And is there are reason to not simple use the full project name rather than an abbreviation?

caroline-gschwend: "nt" = no text. I can change the filenames if you'd rather

gentlecat:
Please use SVG versions of logos. You can find them at https://github.com/metabrainz/metabrainz-logos/tree/master/final-logos.

metabrainz/templates/index/projects.html
gentlecat: I would just put full project name there. Also, names of static files should always be lowercase.

Freso: I think it'd be better to follow the naming scheme from the metabrainz-logos repository (I think that's "ProjectName_no_text.ext"?).

We'll eventually want to do a ["StaticBrainz"](http://tickets.musicbrainz.org/browse/MBH-413) site for holding all our static resources (like images), so the more identically they're named across instances where they're used, the easier it will be to find them and change them to point to staticbrainz.org instead of their own site. :)

metabrainz/templates/index/projects.html
```diff
@@ -0,0 +1,134 @@
...
+{% block title %}Projects - MetaBrainz Foundation{% endblock %}
+
+{% block content %}
+<h1 class="page-title">The MetaBrainz Foundation Projects</h1>
+<div id="projects">
+  
+  <div class="media">
+    <div class="media-left media-middle">
+      <a href="https://musicbrainz.org/">
+        <img class="media-object logo" src="{{ url_for('static', filename='img/projects/musicbrainz.svg') }}" alt="MusicBrainz logo">
```
Freso: Is there a reason you're not using the naming style from metabrainz-logos?

metabrainz/views.py
```diff
@@ -16,8 +16,18 @@ def home():
...
+@index_bp.route('/projects')
+def projects():
+    return render_template('index/projects.html')
 
 
+@index_bp.route('/projects')
+def projects():
+    return render_template('index/projects.html')
+    
+
```
Freso: You'll likely want to remove one of these…

metabrainz/templates/index/projects.html
Freso: Oh, I see @Gentlecat's comment https://github.com/metabrainz/metabrainz.org/pull/208#discussion_r50640384 - I'm not convinced that my argumentation is not proper though. The names are definitely better than they were before. ;)

metabrainz/views.py
gentlecat: Why?

gentlecat: https://www.python.org/dev/peps/pep-0008/#blank-lines

Freso: There were two of
```
+@index_bp.route('/projects')
+def projects():
+    return render_template('index/projects.html')
```

One is now removed.

gentlecat: Oh, I thought you were talking about whitespace. :confused:

Freso: https://github.com/caroline-gschwend/metabrainz.org/commit/b1cb99d642886f5c07e099153bf3646bebc3a7ba

Freso: I wasn't. I know there should be two blank lines here. :wink:

Freso:
Can you make a full-page screenshot and put it up somewhere? (Even here is good.) The text isn't easily readable with the currently code formatting, and I also have no clue how the page currently presents itself.

caroline-gschwend:
screen shot 2016-01-25 at 8 04 26 am
screen shot 2016-01-25 at 8 04 29 am

caroline-gschwend:
screen shot 2016-01-25 at 8 24 27 am

Text justified

caroline-gschwend:
screen shot 2016-01-25 at 9 32 32 am

Freso:
I like this latest layout. :)

Freso:
AFAICT: :shipit:

gentlecat:
I like this layout too.

caroline-gschwend:
I as well like this new layout :)

caroline-gschwend:
screen shot 2016-01-25 at 11 28 06 am

PR #209, opened 2016-01-29
Team page improvements

gentlecat:


Freso:
Maybe the project leader's project should be listed in the same way that the employees' position are?

Freso:
No further comments apart from that and https://github.com/metabrainz/metabrainz.org/pull/210 :)

PR #222, opened 2016-05-22
Change HTTPS-capable http:// links to use https:// instead

Freso:
Also snuck in an update for a wiki.mb.o link to point to mb.o/doc/ instead.


gentlecat:
LGTM

PR #225, opened 2016-06-08
Fix issue with incorrect headers in receipt emails

gentlecat:
*This also fixes several small donation-related issues.*


metabrainz/payments/receipts.py
```diff
@@ -30,11 +30,11 @@ def send_receipt(email, date, amount, name, is_donation, editor_name=None):
             "taxes under section 170 of the Internal Revenue Service code.\n\n"
             "Please save a printed copy of the attached PDF receipt for your records."
         ) % name
-        from_addr = 'donations@' + current_app.config['MAIL_FROM_DOMAIN'],
+        from_addr = 'donations@' + current_app.config['MAIL_FROM_DOMAIN']
```
gentlecat: This was the actual issue. `from_addr` became a tuple, which broke the header generation.

mayhem:
Without actually compiling and test, LGTM

PR #234, opened 2016-07-19
Update logos

gentlecat:
*See http://tickets.musicbrainz.org/browse/MEB-38.*


mayhem:
I don't even know how to review this.

gentlecat:
There's a `Display the rich diff` button next to each file at https://github.com/metabrainz/metabrainz.org/pull/234/files. I guess you could also switch to that branch and see how it looks on the website locally.

mayhem:
I really don't like the improved text. I think you should get others to review this, I'm not for merging this.

gentlecat:
**Examples with large versions (old on the left, updated on the right):**
screenshot 2016-08-15 20 33 27
screenshot 2016-08-15 20 32 08
screenshot 2016-08-15 20 32 14
screenshot 2016-08-15 20 34 14
screenshot 2016-08-15 20 30 22
screenshot 2016-08-15 20 30 27
screenshot 2016-08-15 20 32 45

gentlecat:
Updated version of small logos are already used on https://musicbrainz.org/ and https://critiquebrainz.org/ (see navbars).

uklauer:
As I’ve said repeatedly before, I’m all for fixing the haphazard letter heights (which are plain ugly), so +1.

The new texts are less condensed than the original ones. I’m on the fence about that.

Why the reversed colours for Picard?

mwiencek:
The proportions in the new ones look better and are easier to read to me. Otherwise they aren't much different except for being slightly wider.

gentlecat:
> The new texts are less condensed than the original ones.

Right, that's another thing that makes it looks slightly differently. But that's the same Roboto Black that was used originally.

> Why the reversed colours for Picard?

That's a *bug*. I'll fix it.

gentlecat:
*There was a discussion about this PR and associated ticket during the weekly meeting. See https://chatlogs.metabrainz.org/brainzbot/metabrainz/2016-08-15/?msg=3668733&page=5.*

CatQuest:
fwiw the reversed colours on picard look better to me.

PR #237, opened 2016-11-10
Update sponsors page regarding DWNI

uklauer:
Move DWNI, as former hosting provider, to the “special thanks” section.

Also includes some minor language/typography fixes.


metabrainz/templates/index/sponsors.html
```diff
@@ -39,15 +39,6 @@ <h1 class="page-title">Sponsors</h1>
     </tr>
 
     <tr>
-      <td><img src="{{ url_for('static', filename='img/logos/sponsors/digitalwest.gif') }}"></td>
```
gentlecat: I guess the image also needs to be removed.

PR #238, opened 2016-11-16
MEB-63: Fix sorting of replication packets

gentlecat:
This caused the API to display incorrect number of the latest packet.


mayhem:
🐘 💈 🤐  :shipit:

Issue #77, opened 2015-03-26
Python 3 support


gentlecat:
It would be nice to start using modern version since it's a new project. Right now most of our dependencies support Python 3 except python-memcached, Fabric, and wepay.

* [ ] [python-memcached](https://github.com/linsomniac/python-memcached/) project is working on adding support, it seems like some things are not working yet. See https://github.com/linsomniac/python-memcached/issues/48.
* [ ] [Fabric](http://www.fabfile.org/) doesn't support Python 3 yet. They are working on that as well. See https://github.com/fabric/fabric/issues/1050.
* [x] [WePay Python SDK](https://github.com/wepay/Python-SDK) fails to install with Python 3. I submitted a patch that should fix this problem. See https://github.com/wepay/Python-SDK/pull/18.

*I'll be marking issues that have been fixed.*

tsukanov:
Problem with WePay should be fixed now.

eirnym:
As you look at [roadmap of Fabric](http://www.fabfile.org/en/latest/roadmap.html), it'll become library-oriented with [pyinvoke](https://github.com/pyinvoke/invoke) for task running.

Now you can replace Fabric with pyinvoke without many changes

gentlecat:
Thanks for the tip! I'll look into that.

haypo:
Hi,

I wrote a pull request to finish the python-memcached port to Python 3:
https://github.com/linsomniac/python-memcached/pull/67

Please test it!

Victor

mayhem:
This issue has moved to: http://tickets.musicbrainz.org/browse/MEB-22

PR #240, opened 2016-11-25
MEB-22: Switch to custom Python 3 image

gentlecat:
See http://tickets.metabrainz.org/browse/MEB-22 (and #77). 🐍


gentlecat:
*Still need to do additional tests with donations.*

gentlecat:
Seem to be working.

PR #243, opened 2016-12-07
MEB-8: Catch SMTP exception in the mail sending module

gentlecat:


gentlecat:
Well, it can be difficult to determine what invalid is. We do have formatting restrictions in the form on the website, so users have to provide an email address that is formatted like one. This makes sure that page load doesn't fail even if
* their mail server is unreachable
* our SMTP server is down
* mailbox that they specified doesn't exist

mayhem:
Ok. :shipit:

PR #245, opened 2016-12-10
MEB-50: Redirecting about/customers to /supporters page

dpmittal:
/about/customers link is a dead link and it should be redirected to /supporters link 

Issue Link : https://tickets.metabrainz.org/browse/MEB-50

### DIFF
--- a/metabrainz/views.py
+++ b/metabrainz/views.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, render_template
+from flask import Blueprint, render_template, redirect, url_for
 from metabrainz.model.user import User
 
 index_bp = Blueprint('index', __name__)
@@ -49,3 +49,8 @@ def bad_customers():
 @index_bp.route('/privacy')
 def privacy_policy():
     return render_template('index/privacy.html')
+
+
+@index_bp.route('/about/customers.html')
+def about_customers_redirect():
+    return redirect(url_for('users.supporters_list') , 301)
--- a/metabrainz/views.py
+++ b/metabrainz/views.py
@@ -21,12 +21,12 @@ def about():
 @index_bp.route('/projects')
 def projects():
     return render_template('index/projects.html')
-    
+
 
 @index_bp.route('/team')
 def team():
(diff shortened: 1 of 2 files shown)


metabrainz/views.py
```diff
@@ -49,3 +49,7 @@ def bad_customers():
 @index_bp.route('/privacy')
 def privacy_policy():
     return render_template('index/privacy.html')
+
+@index_bp.route('/about/customers')
```
uklauer: The old URL was `/about/customers.html`, so it seems you are missing the `.html` here.

metabrainz/views.py
```diff
@@ -1,4 +1,4 @@
-from flask import Blueprint, render_template
+from flask import Blueprint, render_template ,redirect , url_for
```
uklauer: The spaces are all over the place here. It should be `foo, bar` – no space before the comma, one space after it, just as in English text.

metabrainz/views.py
```diff
@@ -21,12 +21,12 @@ def about():
...
 def projects():
     return render_template('index/projects.html')
-    
+
 
 @index_bp.route('/team')
 def team():
     return render_template('index/team.html')
-    
+
```
uklauer: These spaces didn’t belong here so it was right to remove them, but that has nothing to do with the redirect issue. Therefore, these indentation fixes should be in a separate commit.

dpmittal:
@uklauer  I have made the changes

uklauer:
I won’t judge whether it is better to remove the indented lines altogether or just the indentation, but in any case, my point about doing it in a separate commit still stands.

gentlecat:
@brainzbot, add to whitelist.

dpmittal:
@uklauer I have fixed the indentation and implemented the redirect feature in two separate commits as you said and @gentlecat I am actually new to flask so I dont know how to write tests at present but I will learn it and will surely send a PR afterwards :)

dpmittal:
@gentlecat I have added the test in a new commit :)

PR #246, opened 2016-12-15
User editing interface updates + logo uploading

gentlecat:


admin/sql/create_tables.sql
```diff
@@ -20,6 +20,7 @@ CREATE TABLE "user" (
   contact_email    CHARACTER VARYING NOT NULL,
   data_usage_desc  TEXT,
   org_name         CHARACTER VARYING,
+  logo_filename    CHARACTER VARYING,
```
mayhem: Why is this needed? Can the filename not be parsed from the org_logo_url?

mayhem: Ah, I see. You've depricated the old logo URL, but still kept it. I think this complicated the code unnecessarily and we should get rid of the old org_logo_url and write a script that moves the legacy supporters into the new format. Then drop the old column.

metabrainz/templates/admin/users/edit.html
```diff
@@ -0,0 +1,240 @@
+{% extends 'admin/master.html' %}
```
mayhem: Was there no old template that did most of this before? Looks like a new file, but we had this feature before and I don't see any files being deleted.

admin/sql/create_tables.sql
gentlecat: Then we can remove it afterwards. I think it would be faster to just move all the logos by hand than to write a script. There aren't as many, I'll do it myself.

PR #247, opened 2016-12-20
User page cleanup

gentlecat:
Similar to the edit page.

### DIFF
--- a/metabrainz/templates/admin/users/details.html
+++ b/metabrainz/templates/admin/users/details.html
@@ -1,8 +1,34 @@
 {% extends 'admin/master.html' %}
 {% block body %}
   <h1>Users</h1>
+  <hr />
 
-  <h3>Details</h3>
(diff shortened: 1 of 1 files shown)


metabrainz/templates/admin/users/details.html
```diff
@@ -13,81 +39,160 @@
...
-          You will be redirected to the next pending user after approving
-          or rejecting this one.
+  <h3>Data access</h3>
+  <div class="row">
+    <div class="form-group col-md-6">
+      <label>State</label>
+      <input class="form-control" type="text" value="{{ user.state }}" readonly>
+      <div style="padding-top: 7px">
+        <em class="text-muted">
+          State determines whether this user has access to the MetaBrainz API.
```
mayhem: "State determines whether this user has access to the MetaBrainz Live Data Feed API."

PR #254, opened 2017-02-07
MEB-80: Switch to another exim container

gentlecat:


consul_config.py.ctmpl
```diff
@@ -59,8 +59,8 @@ LOG_SENTRY = {
 RECAPTCHA_PUBLIC_KEY = '''{{template "KEY" "recaptcha/public_key"}}'''
 RECAPTCHA_PRIVATE_KEY = '''{{template "KEY" "recaptcha/private_key"}}'''
 
-{{if service "metabrainz-mail"}}
-{{with index (service "metabrainz-mail") 0}}
+{{if service "exim-relay"}}
```
zas: This code doesn't handle tags, see https://github.com/metabrainz/musicbrainz-server/pull/455#issue-205910271
The service is named `exim-relay` but it is tagged with the domain, here `metabrainz-org`.
Also it may be convenient to be able to use an env var to configure the actual name of the service, as done in https://github.com/metabrainz/musicbrainz-server/pull/455/files

See also https://github.com/metabrainz/docker-server-configs/blob/master/scripts/services.sh#L278

PR #256, opened 2017-02-16
Update account-type.html

eebigger:
Updates to the Verbiage of Account-type webpage.

### DIFF
--- a/metabrainz/templates/users/account-type.html
+++ b/metabrainz/templates/users/account-type.html
@@ -6,7 +6,7 @@
   <h1 class="page-title">{{ _('Support us') }}</h1>
 
   <p>
-    {{ _('If you rely on data from MetaBrainz, please remember that *we* rely on contributions 
+    {{ _('If you rely on data from MetaBrainz, please remember that we rely on contributions 
     to keep providing it. For commercial use of our data or to get access to the
     <a href="%(mb_data_feed)s">MusicBrainz Live Data Feed</a>,
     please choose between non-commercial and commercial use below. You will need a
--- a/metabrainz/templates/users/account-type.html
+++ b/metabrainz/templates/users/account-type.html
@@ -6,25 +6,24 @@
   <h1 class="page-title">{{ _('Support us') }}</h1>
 
   <p>
-    {{ _('For commercial use of our data or to get access to the
+    {{ _('If you rely on data from MetaBrainz, please remember that *we* rely on contributions 
+    to keep providing it. For commercial use of our data or to get access to the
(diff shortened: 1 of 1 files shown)


metabrainz/templates/users/account-type.html
```diff
@@ -6,25 +6,24 @@
   <h1 class="page-title">{{ _('Support us') }}</h1>
 
   <p>
-    {{ _('For commercial use of our data or to get access to the
+    {{ _('If you rely on data from MetaBrainz, please remember that *we* rely on contributions
```
mayhem: I think the emphasis on *we* is not quite apt. First, it should bold or italic, not in * *, since * * is a way to make things bold in plain text, but we're on a web page. Second, I think that emphasis is a bit unprofessional -- I would just nuke it.

eebigger: +1

metabrainz/templates/users/account-type.html
```diff
@@ -6,25 +6,24 @@
...
     please choose between non-commercial and commercial use below. You will need a
     <a href="%(mb_register_url)s">MusicBrainz account</a> with a verified email address to sign-up.',
     mb_data_feed='https://musicbrainz.org/doc/Live_Data_Feed', mb_register_url='https://musicbrainz.org/register') }}
   </p>
   <p>
-    {{ _('<strong>Open source developers and non-profits</strong>: Please sign up as a
-    commercial user using the free non-profit tier below. This allows us to give you credit
-    for your open source project or your non-profit.') }}
-  </p>
+
```
gentlecat: ✂️

metabrainz/templates/users/account-type.html
```diff
@@ -6,25 +6,24 @@
...
   <p>
-    {{ _('For commercial use of our data or to get access to the
+    {{ _('If you rely on data from MetaBrainz, please remember that we rely on contributions 
+    to keep providing it. For commercial use of our data or to get access to the
     <a href="%(mb_data_feed)s">MusicBrainz Live Data Feed</a>,
     please choose between non-commercial and commercial use below. You will need a
     <a href="%(mb_register_url)s">MusicBrainz account</a> with a verified email address to sign-up.',
     mb_data_feed='https://musicbrainz.org/doc/Live_Data_Feed', mb_register_url='https://musicbrainz.org/register') }}
   </p>
   <p>
```
gentlecat: The opening tag needs to be removed too.

mayhem: Opening tag?

gentlecat: `</p>` below is removed, so this `<p>` should be too.

PR #258, opened 2017-04-05
Prevent literal <em> from being displayed

navap:
I didn't test this.
I haven't done this in years.
I might have just broken everything.

Buyer beware.


gentlecat:
@brainzbot, add to whitelist.

PR #264, opened 2017-04-18
MEB-90: Add support for donating in EUR using PayPal

gentlecat:


metabrainz/model/payment.py
```diff
@@ -160,14 +166,29 @@ def process_paypal_ipn(cls, form):
...
             return
 
+        account_ids = current_app.config['PAYPAL_ACCOUNT_IDS']  # "currency => account" mapping
+
+        if form['mc_currency'].lower() not in SUPPORTED_CURRENCIES:
+            logging.warning("Unsupported currency: ", form['mc_currency'])
+            return
+
+        # Checking if payment was sent to the right account depending on the currency
+        if form['mc_currency'].upper() in account_ids:
```
mayhem: Why would this ever happen?

metabrainz/templates/payments/payment.html
```diff
@@ -219,7 +219,7 @@ <h1 class="page-title">{{ _('Make a Payment') }}</h1>
 
         {# Info about variables can be found at https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/Appx_websitestandard_htmlvariables/ #}
 
-        $('<input>').attr({type: 'hidden', name: 'business', value: '{{ config.PAYPAL_PRIMARY_EMAIL }}'}).appendTo(form);
+        $('<input>').attr({type: 'hidden', name: 'business', value: '{{ config['PAYPAL_ACCOUNT_IDS']['USD'] }}'}).appendTo(form);
```
mayhem: Why is this hard coded to USD?

gentlecat: I haven't changed "payments" yet. Just donations.

metabrainz/model/payment.py
gentlecat: When money is sent to an account for another currency. If that's too redundant, I can remove the check.

metabrainz/templates/payments/payment.html
mayhem: Ah, I saw the [WIP] on the title of the PR go away, so I thought it was done...

metabrainz/model/payment.py
mayhem: Ideally we should try and avoid this happening in the first place -- is there reason to think that this might still occur?

metabrainz/templates/payments/payment.html
gentlecat: I split it into two.

metabrainz/model/payment.py
gentlecat: If there is a bug in the frontend.

gentlecat:
> We'll wait to deploy until the latter half is done, yes?

Not necessarily. This should work just fine.

I'm currently fixing database creation issues that you pointed out before.

PR #266, opened 2017-04-19
Expose PostgreSQL port in development

gentlecat:


gentlecat:
I'd prefer to still make PRs even if it might be a small change. I want to make sure that tests still pass in Jenkins, and maybe someone would have input on any kinds of changes. Everyone makes mistakes and there are always improvements to be had. Not necessary to review everything, I understand we don't have enough people.

mayhem:
Got it. In this case, can you mark the reviews that don't strictly need reviews as such?

gentlecat:
Sure, we can come up with some kind of label like `minor-change` and put them on less important PRs. I also usually explicitly assign reviewers when review is needed.

mayhem:
Good to know, and I think I'd also like the minor-change label, please.

PR #273, opened 2017-04-27
MEB-75: Add filters to show only payments or donations in the admin view

gentlecat:
screenshot 2017-04-27 17 33 14


gentlecat:
@eebigger, does that look good?

mayhem:
k

eebigger:
I like this! this is a great solution for sorting between the two. I would implement this asap while I marinate on other possible improvements!

PR #274, opened 2017-04-27
Allow to specify tag for images pushed to the Docker Hub

gentlecat:
https://docs.docker.com/engine/reference/commandline/push/

### DIFF
--- a/docker/push.sh
+++ b/docker/push.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
+#
+# Build image from the currently checked out version of the MetaBrainz website
+# and push it to the Docker Hub, with an optional tag (by default "latest").
+#
+# Usage:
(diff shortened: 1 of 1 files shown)


docker/push.sh
```diff
@@ -1,6 +1,16 @@
...
+# and push it to the Docker Hub, with an optional tag (by default "latest").
+#
+# Usage:
+#   $ ./push.sh [tag]
 
 cd "$(dirname "${BASH_SOURCE[0]}")/../"
 
-docker build -t metabrainz/metabrainz .
-docker push metabrainz/metabrainz
+TAG_PART=""
```
mwiencek: Alternatively, just `TAG_PART=${1:-latest}` to provide a default value (but make sure to always include the `:` in the commands below if you do that).

PR #280, opened 2017-05-08
Update About Page ORG-8

eebigger:
Updated verbiage with new links. Moved logo to top of page.

Unsure if some of my changes are in correct format. Would like to review and learn from errors.

### DIFF
--- a/metabrainz/templates/index/about.html
+++ b/metabrainz/templates/index/about.html
@@ -3,42 +3,44 @@
 {% block title %}{{ _('About') }} - MetaBrainz Foundation{% endblock %}
 
 {% block content %}
-  <h1 class="page-title">{{ _('About the MetaBrainz Foundation') }}</h1>
+  <h1 class="page-title"><img src="{{ url_for('static', filename='img/logos/metabrainz.svg') }}" class="hidden-xs"></h1>
+
+     <h2 class="page-title">{{ _('About the MetaBrainz Foundation') }}</h2>
 
   <div class="row">
     <div class="col-sm-9">
       <p>
-        {{ _('The MetaBrainz Foundation is a 501(c)(3) tax-exempt non-profit founded in
-        December 2004 and based in San Luis Obispo, California, that believes in
-        free, open access to data.') }}
+        {{ _('The MetaBrainz Foundation is a global community creating and maintaining 
+        an open encyclopedia of music and arts metadata. Beginning as a resource for music 
+        collectors we currently establish collaborations between volunteers, artists, and 
+        industry leaders. Since it’s founding it has grown into a global community of 
+        <a href="%(editors_url)s">hundreds of thousands of editors</a> 
+        collaborating together to build and maintain the MetaBrainz comprehensive datasets.') }}
       </p>
       <p>
-        {{ _('The MetaBrainz Foundation has been set up to build community maintained
-        databases and make them available in the public domain or under Creative
-        Commons licenses. Our data is mostly gathered by volunteers, and verified
-        by a voting system to make sure it is consistent and correct. All
-        non-commercial use is free, but commercial users are asked to pay for
-        regular updates of the data in order to help fund the project. We
-        encourage all data users to contribute to the data gathering process so
-        that our data can be as comprehensive as possible.') }}
+        {{ _('The foundation works within a balanced ecosystem of contributions to the datasets 
+        and community and the utilization of the datasets. We encourage all data users to participate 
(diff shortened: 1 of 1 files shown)


metabrainz/templates/index/about.html
```diff
@@ -3,42 +3,44 @@
...
     <div class="col-sm-9">
       <p>
-        {{ _('The MetaBrainz Foundation is a 501(c)(3) tax-exempt non-profit founded in
-        December 2004 and based in San Luis Obispo, California, that believes in
-        free, open access to data.') }}
+        {{ _('The MetaBrainz Foundation is a global community creating and maintaining 
+        an open encyclopedia of music and arts metadata. Beginning as a resource for music 
+        collectors we currently establish collaborations between volunteers, artists, and 
+        industry leaders. Since it’s founding it has grown into a global community of 
+        <a href="%(editors_url)s">hundreds of thousands of editors</a>
```
gentlecat: All the placeholders like `%(editors_url)s` need to have a value defined to be filled in when the template is rendered. This can be done by passing them as an argument to the `_` (aka `gettext`) function that wraps the whole string.

```python
{{ _('A string of text with some placeholders inside. Like %(this_example)s...',
     this_example="this string") }}
```

metabrainz/templates/index/about.html
```diff
@@ -3,42 +3,44 @@
...
-        by a voting system to make sure it is consistent and correct. All
-        non-commercial use is free, but commercial users are asked to pay for
-        regular updates of the data in order to help fund the project. We
-        encourage all data users to contribute to the data gathering process so
-        that our data can be as comprehensive as possible.') }}
+        {{ _('The foundation works within a balanced ecosystem of contributions to the datasets 
+        and community and the utilization of the datasets. We encourage all data users to participate 
+        in the data gathering, editing and peer review processes so that our data can be as comprehensive
+        as possible. Each entity is assigned a permanent 
+        <a href="%(MusicBrainz_Identifier_url)s">MBID (MusicBrainz Identifier)</a>
```
gentlecat: It's also a good idea to keep all names for placeholder values lowercase since:

1. They are basically python/jinja variable names.
2. It's easier for localisers not to worry about what case these names are in (since they are case sensitive) and just assume that they are lowercase by default.

metabrainz/templates/index/about.html
```diff
@@ -3,42 +3,44 @@
...
-        financial_reports_url=url_for('financial_reports.index'),
-        supporters_url=url_for('users.supporters_list'),
-        sponsors_url=url_for('index.sponsors')) }}
+        {{ _('The foundation was founded in 2004 by Robert Kaye, in San Luis Obispo, California and is a 
+        501(c)(3) tax-exempt non-profit. <a href="%(supporters_url)s">Commercial users</a> are asked to contribute 
+        financially in order to help fund our family of projects, and give back to our community to keep our 
+        ecosystem healthy. Non-commercial use of our data remains free, and we are committed to keeping it free to utilize. 
+        We are also supported by donations and a number of <a href="%(sponsors_url)s">sponsors</a> who provide 
+        funds in order for the foundation to continue to accomplish various project goals. The MetaBrainz 
+        Foundation believes in<a href="%(financial_reports_url)s">transparent finances,</a> which are published quarterly.
```
gentlecat: Make sure to define link values (for placeholders), close the function that was opened with `_(`, and close the template block that was opened with `{{`.

metabrainz/templates/index/about.html
eebigger: var string = "hundreds of thousands of editors";
var result = string link("https://musicbrainz.org/statistics/editors"); 

So something like below does the same as above? 

{{ _( %(hundreds of thousands of editors)s, hundreds of thousands of editors="https://musicbrainz.org/statistics/editors") }}

metabrainz/templates/index/about.html
eebigger: excellent

metabrainz/templates/index/about.html
gentlecat: There are some issues with the syntax. It looks kind of like JavaScript, but templates are written in [Jinja](http://jinja.pocoo.org/docs/2.9/templates/). The main problem with the example below is that variable name can't have spaces in it. It's also a bit too long. :)

gentlecat:
I merged this separately and fixed issues in 517c98a9f737e5b855b2366f503758051f205e95.

PR #290, opened 2017-12-01
Right-align finance values in yearly reports, fix minor mistakes in 2015 report

LeoVerto:
Rob wanted the dollar values to be right-aligned and there were some other minor mistakes in the 2015 report.

The diff reports the older reports as having been completely rewritten as for some reason they are a single line of html.
You can run "git diff --word-diff=color a493c3a" to see what's actually been changed (I added the "finances" class to the relevant tables).


mayhem:
@brainzbot, add to whitelist

mayhem:
Thank you!

PR #291, opened 2017-12-04
Only use consul config in production

mayhem:
Only read the consul configuration file and nothing else if we're running in production. The steps that CustomFlask takes to find a configuration file can hide problems in production. A race condition to find the consul config file and this issue caused an hour long metabrainz.org outage, so I'm keen to really ensure this is fixed.

### DIFF
--- a/metabrainz/__init__.py
+++ b/metabrainz/__init__.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from time import sleep
 
 from brainzutils.flask import CustomFlask
@@ -7,6 +8,8 @@ from flask import send_from_directory, request
 # Check to see if we're running under a docker deployment. If so, don't second guess
 # the config file setup and just wait for the correct configuration to be generated.
 deploy_env = os.environ.get('DEPLOY_ENV', '')
+
+CONSUL_CONFIG_FILE_RETRY_COUNT = 10 
 
 def create_app(config_path=None):
     app = CustomFlask(
@@ -21,9 +24,14 @@ def create_app(config_path=None):
         print("Running in production!");
         consul_config = os.path.join( os.path.dirname(os.path.realpath(__file__)), 
             '..', 'consul_config.py')
-        while not os.path.exists(consul_config):
-            print("No configuration file generated yet. Sleeping...");
-            sleep(2)
+
+        for i in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
+            if not os.path.exists(consul_config):
+                sleep(1)
+                    
+        if not os.path.exists(consul_config):
+            print("No configuration file generated yet. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT);
+            sys.exit(-1)
 
         app.config.from_pyfile(consul_config, silent=True)
 
--- a/metabrainz/__init__.py
+++ b/metabrainz/__init__.py
@@ -15,8 +15,8 @@ def create_app(config_path=None):
         use_debug_toolbar=True,
     )
 
-    # Load configuration files
-    # If we're running under a docker deployment, only use the consul configuration.
+    # Load configuration files: If we're running under a docker deployment, wait until 
+    # the consul configuration is available.
     if deploy_env:
         print("Running in production!");
         consul_config = os.path.join( os.path.dirname(os.path.realpath(__file__)), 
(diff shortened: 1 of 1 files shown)


metabrainz/__init__.py
```diff
@@ -10,21 +15,30 @@ def create_app(config_path=None):
...
+        consul_config = os.path.join( os.path.dirname(os.path.realpath(__file__)), 
+            '..', 'consul_config.py')
+        while not os.path.exists(consul_config):
+            print("No configuration file generated yet. Sleeping...");
+            sleep(2)
+
+        app.config.from_pyfile(consul_config, silent=True)
+
+    else:
+        # Otherwise, do our best to find configu files
```
zas: Minor typo: 'configu' -> 'config'

metabrainz/__init__.py
```diff
@@ -1,7 +1,12 @@
+import os
+from time import sleep
+
 from brainzutils.flask import CustomFlask
 from flask import send_from_directory, request
-import os
 
+# Check to see if we're running under a docker deployment. If so, don't second guess
+# the config file setup and just wait for the correct configuration to be generated.
+deploy_env = os.environ.get('DEPLOY_ENV', '')
```
zas: Other projects are using 'prod', 'beta', 'test' as deploy env names.
If i understand empty = prod here ?

metabrainz/__init__.py
```diff
@@ -10,21 +15,30 @@ def create_app(config_path=None):
...
+    # If we're running under a docker deployment, only use the consul configuration.
+    if deploy_env:
+        print("Running in production!");
+        consul_config = os.path.join( os.path.dirname(os.path.realpath(__file__)), 
+            '..', 'consul_config.py')
+        while not os.path.exists(consul_config):
+            print("No configuration file generated yet. Sleeping...");
+            sleep(2)
+
+        app.config.from_pyfile(consul_config, silent=True)
```
alastair: If you're going to wait for this file, there's probably no need for `silent=True` - removing this check will make the app explicitly fail if you get here and it doesn't exist, which is probably a good idea

metabrainz/__init__.py
```diff
@@ -10,21 +15,30 @@ def create_app(config_path=None):
...
+
+    else:
+        # Otherwise, do our best to find configu files
+        app.config.from_pyfile(os.path.join(
+            os.path.dirname(os.path.realpath(__file__)),
+            '..', 'default_config.py'
```
alastair: I don't know how config files in production work here - does consul_config contain all settings that exist in default_config, or does it just override a few things? If it overrides only some settings you should always do this one, and then optionally  load consul/custom config in your if block

metabrainz/__init__.py
```diff
@@ -10,21 +15,30 @@ def create_app(config_path=None):
...
+        ))
+        app.config.from_pyfile(os.path.join(
+            os.path.dirname(os.path.realpath(__file__)),
```
zas: It would be worth to cache `os.path.dirname(os.path.realpath(__file__))`, repeated 3 times.

metabrainz/__init__.py
mayhem: It provides everything needed to run the service. This is the template file that is filled out with actual values:

https://github.com/metabrainz/metabrainz.org/blob/master/consul_config.py.ctmpl

No other config files should be used or values overridden.

metabrainz/__init__.py
```diff
@@ -10,21 +15,30 @@ def create_app(config_path=None):
...
-        app.config.from_pyfile(config_path)
+    # Load configuration files
+    # If we're running under a docker deployment, only use the consul configuration.
+    if deploy_env:
+        print("Running in production!");
+        consul_config = os.path.join( os.path.dirname(os.path.realpath(__file__)), 
+            '..', 'consul_config.py')
+        while not os.path.exists(consul_config):
+            print("No configuration file generated yet. Sleeping...");
+            sleep(2)
```
zas: `sleep(1)` is prolly enough (since it loops anyway).
I would not emit a message on first pass but only after a number of loops, to prevent cluttering the log file for nothing.
Also i usually limit the number of loops: emit a critical message and exit if the file didn't appear after a long time.
It prevents silent failures, in this case, something is definitively broken.

metabrainz/__init__.py
mayhem: Yes, it will restart the service, but I am not sure of all of the semantics in place, since our current setup is unclear to me and not properly documented. But keeping silent=True and explicitly checking I can guarantee what is going to happen and that way I can guarantee that it is going to do the right thing. I am ok with improving this once the underlying issue has been solved. For now, there is little harm in doing it this way.

metabrainz/__init__.py
alastair: There are many items in default_config that don't exist in the consul_config template:
https://github.com/metabrainz/metabrainz.org/blob/7738c77b7ffb4eac336a399878b28255dbd79fe9/default_config.py#L16-L17

mayhem: Ugh. Good catch! That makes this solution a lot less clean. :(

PR #292, opened 2017-12-05
Start uwsgi with consul-template's exec mode

mwiencek:
The new `metabrainz/python:3.6-1` image is based on `metabrainz/consul-template-base:v0.18.5-2`.

I removed some old, unnecessary files under docker-helpers.

I haven't been able to test this yet. Can we use https://test.metabrainz.org/ to do that?


mayhem:
This PR, merged into the startup-logging branch is now running on https://test.metabrainz.org/

PR #293, opened 2017-12-07
MEB-98: Add logging of config values and git commit on init

paramsingh:
Also, removed the `runserver` command from the dockerfile and moved it into `docker-compose.yml` for consistency with LB and MsB etc.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -6,6 +6,7 @@ services:
     build:
       context: ..
       dockerfile: ./docker/Dockerfile.dev
+    command: python /code/manage.py runserver -h 0.0.0.0 -p 80
     volumes:
       - ../data/replication_packets:/data/replication_packets
       - ../data/json_dumps:/data/json_dumps
--- a/metabrainz/utils.py
+++ b/metabrainz/utils.py
@@ -1,6 +1,8 @@
 from urllib.parse import urlparse, parse_qsl, urlunparse, urlencode
-import string
+
 import random
+import string
+import subprocess
 
 
 def reformat_datetime(value, format='%x %X %Z'):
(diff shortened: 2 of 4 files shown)


mayhem:
@brainzbot please test this

docker/docker-compose.dev.yml
```diff
@@ -6,6 +6,7 @@ services:
     build:
       context: ..
       dockerfile: ./docker/Dockerfile.dev
+    command: python /code/manage.py runserver -h 0.0.0.0 -p 80
```
mayhem: curious, why did you move this?

metabrainz/utils.py
```diff
@@ -1,6 +1,8 @@
 from urllib.parse import urlparse, parse_qsl, urlunparse, urlencode
-import string
+
 import random
+import string
```
mayhem: Is this needed?

mayhem:
@brainzbot please add this to the whitelist

mayhem:
@brainzbot, add to whitelist

mayhem:
@brainzbot, you're useless twat.

docker/docker-compose.dev.yml
paramsingh: @mayhem, I was looking at the Dockerfile to see why debug messages didn't seem to be getting printed, so I tried moving the command here. Anyways, doing this in the compose file looked better to me (and it is consistent with how we do it in LB and MsB etc.)

metabrainz/utils.py
paramsingh: I moved the imports around to make them alphabetical. The import is  used in a the [generate_string](https://github.com/metabrainz/metabrainz.org/blob/81fc69a531080928f41657f6a755f4827ddd4820/metabrainz/utils.py#L14) function.

PR #294, opened 2017-12-11
Improve the startup logging and fix the broken git version logging

mayhem:
as it says on the tin..

### DIFF
--- a/metabrainz/__init__.py
+++ b/metabrainz/__init__.py
@@ -61,7 +61,7 @@ def create_app(config_path=None):
             git_version = f.read()
         print('Running on git commit %s' % git_version.strip())
     except IOError:
-        print('Unable to retrieve git commit. Use docker/push.sh to push images for production.")
+        print("Unable to retrieve git commit. Use docker/push.sh to push images for production.")
 
     print('Configuration values are as follows: ')
     print(pprint.pformat(app.config, indent=4))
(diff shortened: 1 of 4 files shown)


metabrainz/__init__.py
```diff
@@ -56,23 +55,23 @@ def create_app(config_path=None):
 
         app.config.from_pyfile(consul_config, silent=True)
 
+    # Printing out some debug values such as config and git commit
+    try:
+        with open(".git-version") as f:
+            git_version = f.read()
+        print('Running on git commit %s' % git_version.strip())
+    except subprocess.CalledProcessError as e:
```
paramsingh: This error will probably not be thrown now.

metabrainz/__init__.py
```diff
@@ -56,23 +55,23 @@ def create_app(config_path=None):
 
         app.config.from_pyfile(consul_config, silent=True)
 
+    # Printing out some debug values such as config and git commit
+    try:
+        with open(".git-version") as f:
```
paramsingh: We could do this in `utils.get_git_commit`.

metabrainz/__init__.py
mayhem: Oh, good point.

metabrainz/__init__.py
mayhem: Remind me, was this added for just this case or was this an existing function that works in other contexts?

mayhem:
An env variable would make it available to the Dockerfile, but not the running container in production. The Dockerfile would need to write it to a file, so its not much of a difference.

metabrainz/__init__.py
paramsingh: No, just added here, so could be removed easily.

mayhem: Done!

PR #301, opened 2018-04-07
Document incremental JSON dumps API, remove daily/weekly packets

mwiencek:
json-dump-api

The daily/weekly packets haven't been updating since some time in 2016, and I recall @mayhem giving the OK to only support hourly ones, but lemme know if you'd still like to document/include them in the API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,12 +54,6 @@ It must have the following structure:
 ```
 ./data/replication_packets/
     - hourly replication packets
-
-./data/replication_packets/daily/
-    - daily replication packets
-
-./data/replication_packets/weekly/
-    - weekly replication packets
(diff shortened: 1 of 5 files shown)


mwiencek:
I added two more commits:

In https://github.com/metabrainz/metabrainz.org/pull/301/commits/a804574a5673088a5d727222eba2ab1abb32d1a2 I completed the JSON dumps API by adding an endpoint for the .asc signatures, and added some tests.

In https://github.com/metabrainz/metabrainz.org/pull/301/commits/35731bad1cf43f520373110db655dc77a48dc895 I fixed the instructions for getting the replication packet signatures, which were incorrect.

PR #318, opened 2019-04-22
Create a new stats view that shows the most active IP addresses

mayhem:
Adds a fairly simple new admin view to see who the top IP addresses are in the last week.

### DIFF
--- a/metabrainz/__init__.py
+++ b/metabrainz/__init__.py
@@ -20,6 +20,9 @@ def create_app(debug=None, config_path = None):
         use_flask_uuid=True,
     )
 
+    # get ride of some really pesky warning
+    app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
+
     print("Starting metabrainz service with %s environment." % deploy_env);
 
     # This is used to run tests, but not for dev or deployment
--- a/metabrainz/__init__.py
+++ b/metabrainz/__init__.py
@@ -20,7 +20,7 @@ def create_app(debug=None, config_path = None):
         use_flask_uuid=True,
     )
 
-    # get ride of some really pesky warning
+    # get rid of some really pesky warning
     app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
 
     print("Starting metabrainz service with %s environment." % deploy_env);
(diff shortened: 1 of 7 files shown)


metabrainz/model/access_log_test.py
```diff
@@ -0,0 +1,48 @@
...
+            contact_name = "Mr. Commercial",
+            contact_email = "testc@musicbrainz.org",
+            data_usage_desc = "poop!",
+            org_desc = "foo!"
+        )
+        user.set_state(STATE_ACTIVE)
+        token = user.generate_token() 
+        AccessLog.create_record(token, "10.1.1.59")
+
+        non_commercial, commercial = AccessLog().top_ips()
```
alastair: doesn't need `()` after AccessLog because top_ips is a classmethod

metabrainz/__init__.py
```diff
@@ -20,6 +20,9 @@ def create_app(debug=None, config_path = None):
         use_flask_uuid=True,
     )
 
+    # get rid of some really pesky warning
+    app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
```
alastair: maybe better to go in the config file/config file template instead of create_app()?

metabrainz/admin/views.py
```diff
@@ -241,6 +245,54 @@ def overview(self):
...
+        except Exception as err:
+            return None
+
+
+    @staticmethod
+    def lookup_ips(users):
+        """ Try to lookup and cache as many reverse DNS as possible in a window of time """
+
+        data = []
+        timeout = time.time() + StatsView.IP_ADDR_TIMEOUT
```
alastair: if this is python3 you can use time.monotonic(): https://docs.python.org/3/library/time.html#time.monotonic

metabrainz/model/access_log.py
```diff
@@ -147,3 +147,43 @@ def top_downloaders(cls, limit=None):
...
+            list for commercial users and another list for non-commercial users who
+            are not in good standing. good standing for non-commercial users means
+            that they are verified good non-commercial users.
+
+        Args:
+            days: Number of past days to include in the query
+            limit: Max number of items to return.
+
+        Returns:
+            List of <User, request count> pairs
```
alastair: not quite true? should be a tuple of these pairs with non-commercial first and then commercial

metabrainz/__init__.py
mayhem: This is one of those annoying messages that makes test runs hard to read -- I dont want to clutter the config files more than they already are. I'll make a note to remove this in a year, when hopefully it isn't needed anymore.

PR #319, opened 2019-04-24
Add top token stats

mayhem:
The same as the previous PR, but for top tokens. Also add test.sh and this version uses column names and not column indexes.

### DIFF
--- a/metabrainz/model/access_log.py
+++ b/metabrainz/model/access_log.py
@@ -147,3 +147,44 @@ class AccessLog(db.Model):
         if limit:
             query = query.limit(limit)
         return query.all()
+
+    @classmethod
+    def top_ips(cls, days=7, limit=None):
+        """
+            Generates two list of most active ip addresses in the last days. One
+            list for commercial users and another list for non-commercial users who
+            are not in good standing. good standing for non-commercial users means
+            that they are verified good non-commercial users.
+
+        Args:
+            days: Number of past days to include in the query
+            limit: Max number of items to return.
+
+        Returns:
+            Tuple of (non_commercial, commercial) lists of [ip_address, token, musicbrainz_id, user_id, contact_name, contact_email]
+
+        """
+        query = db.session.query(AccessLog).join(Token).join(User) \
+            .with_entities(AccessLog.ip_address, AccessLog.token, User.musicbrainz_id, User.id, User.contact_name, User.contact_email) \
+            .filter(User.is_commercial == False) \
+            .filter(cls.timestamp > datetime.now() - timedelta(days=days)) \
+            .filter(User.good_standing != True) \
+            .add_columns(func.count("AccessLog.*").label("count")) \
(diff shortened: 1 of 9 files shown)


metabrainz/model/access_log.py
```diff
@@ -188,3 +188,45 @@ def top_ips(cls, days=7, limit=None):
...
+            list for commercial users and another list for non-commercial users who
+            are not in good standing. good standing for non-commercial users means
+            that they are verified good non-commercial users.
+
+        Args:
+            days: Number of past days to include in the query
+            limit: Max number of items to return.
+
+        Returns:
+            Tuple of (non_commercial, commercial) lists of [ip_address, token, musicbrainz_id, user_id, contact_name, contact_email]
```
paramsingh: This function doesn't have ip addresses, right?

metabrainz/templates/admin/stats/top-tokens.html
```diff
@@ -0,0 +1,41 @@
...
+       Users in good standing are not listed here.
+    </p>
+    {{ user_list(non_commercial) }}
+
+    <h2>Commercial users</h2>
+    {{ user_list(commercial) }}
+
+{% endblock %}
+
+{% macro user_list(users) -%}
```
paramsingh: Could we reuse the macro from the other template in the previous PR?

mayhem: We can't; there is one columns less in this view.

metabrainz/model/access_log.py
```diff
@@ -147,3 +147,86 @@ def top_downloaders(cls, limit=None):
...
             query = query.limit(limit)
         return query.all()
+
+    @classmethod
+    def top_ips(cls, days=7, limit=None):
+        """
+            Generates two list of most active ip addresses in the last days. One
+            list for commercial users and another list for non-commercial users who
+            are not in good standing. good standing for non-commercial users means
+            that they are verified good non-commercial users.
```
reosarevok: Generates two list**s** of **the** most active **IP** addresses in the last days. One
            list for commercial users and another list for non-commercial users who
            are not in good standing. **G**ood standing for non-commercial users means
            that they are verified good non-commercial users.

metabrainz/model/access_log.py
```diff
@@ -147,3 +147,86 @@ def top_downloaders(cls, limit=None):
...
+        return (non_commercial, commercial)
+
+
+    @classmethod
+    def top_tokens(cls, days=7, limit=None):
+        """
+            Generates two list of most active token in the last days. One
+            list for commercial users and another list for non-commercial users who
+            are not in good standing. good standing for non-commercial users means
+            that they are verified good non-commercial users.
```
reosarevok: Generates two list**s** of **the** most active token**s** in the last days. One
            list for commercial users and another list for non-commercial users who
            are not in good standing. **G**ood standing for non-commercial users means
            that they are verified good non-commercial users.

PR #321, opened 2019-05-08
Add HTML elements to view 1, 7, 30 views of the top reports

mayhem:
Not super pretty but the target audience is limited to 1 person. :) :)

### DIFF
--- a/metabrainz/admin/views.py
+++ b/metabrainz/admin/views.py
@@ -281,7 +281,14 @@ class StatsView(AdminBaseView):
 
     @expose('/top-ips/')
     def top_ips(self):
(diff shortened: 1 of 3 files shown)


metabrainz/admin/views.py
```diff
@@ -281,25 +281,41 @@ def lookup_ips(users):
 
     @expose('/top-ips/')
     def top_ips(self):
-        non_commercial, commercial = AccessLog.top_ips(limit=100, days=7)
+
+        days = 7
```
paramsingh: Maybe this could be shared in a single function?

mayhem: I'll do that for the next round to changes -- there will be quite a few more reports coming.

PR #324, opened 2019-07-08
Add quickbooks invoicing support

mayhem:
This PR adds support for adding invoices to QuickBooks. Sadly we keep two database about our customers (one in QuickBooks and one in meb.org). This PR only works with the customers and invoices that are in QB.

There is one hideous function in this PR that is rather long; breaking it up would make life worse in the grand scheme of things, so let's leave this function as is.

This code has now been tested for at least 3 cycles, so I'm pretty happy with it.


mayhem:
An no, this amazingly complex beast has no tests. At least there is only one target customer: Me. if it breaks, it is my problem to fix and I am the only person affected by it.

paramsingh:
Tests seem to be failing because of some syntaxerrors.

metabrainz/__init__.py
```diff
@@ -76,6 +78,9 @@ def create_app(debug=None, config_path = None):
         sentry_config=app.config.get('LOG_SENTRY'),
     )
 
+    # Don't intercept redirects, which is fucking annoying
```
paramsingh: We put this inside config.py.sample and consul_config.py.ctmpl instead of code.

metabrainz/admin/quickbooks/quickbooks.py
```diff
@@ -0,0 +1,34 @@
...
+from quickbooks import QuickBooks
+from flask import request, current_app
+
+def init(app):
+    '''
+    Create global auth client that manages QuickBooks sessions.
+    '''
+
+    if not app.config["QUICKBOOKS_CLIENT_ID"]:
+        app.logger.warn("No QUICKBOOKS_CLIENT_ID specified, not setting up QB invoice feature.")
```
paramsingh: What happens to the admin views when this is not specified? we still register the views, do they ISE? Maybe we shouldn't register them?

metabrainz/admin/quickbooks/views.py
```diff
@@ -0,0 +1,378 @@
...
+
+            invoices.append( { 'customer' : customer_id, 'begin' : begin_date, 'end' : end_date, 'base_invoice' : base_invoice, 'qty' : qty, 'price' : price })
+            customer += 1 
+
+        # setup the access nonsense again
+        access_token = session.get('access_token', None)
+        refresh_token = session.get('refresh_token', None)
+        realm = session.get('realm', None)
+
+        if not session['access_token']:
```
paramsingh: this should probably be `if not access_token`, since we get it 4 lines above?

paramsingh: what happens when we have an access token but no realm or refresh token? Is that possible?

mayhem: It isn't possible from what I know.

mayhem:
Yes, a test setup of this would be massive amounts of work. Not worth it given our shitty target audience. :)

PR #326, opened 2019-07-13
Upgrade flask admin to fix SEC issue

paramsingh:
GitHub said there wasn't a fix, but looks like 1.5.3 fixes it (https://flask-admin.readthedocs.io/en/latest/changelog/#id1)


mayhem:
Deployed, thanks!

PR #342, opened 2020-04-18
Remove Param from team page

paramsingh:
As it says on the tin, not contracting now, so should be removed.

:sadpanda:


mayhem:
I've reworded your section slightly and moved you to the volunteer section. Not lettting you go that easily! :)

PR #343, opened 2020-04-28
Add current supporters page

mayhem:
I've added a new view that shows our active customers and the amount they have pledged and then the total for all. This view shows what our monthly income should be.

### DIFF
--- a/metabrainz/admin/views_test.py
+++ b/metabrainz/admin/views_test.py
@@ -18,3 +18,6 @@ class AdminViewsTestCase(FlaskTestCase):
 
     def test_statsview_top_ips(self):
         self.assertStatus(self.client.get(url_for('statsview.top_ips')), 302)
+
+    def test_statsview_supporters(self):
+        self.assertStatus(self.client.get(url_for('statsview.supporters')), 302)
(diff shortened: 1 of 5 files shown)


metabrainz/templates/admin/stats/supporters.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'admin/master.html' %}
+{% block body %}
+  <h1>Active supporters</h1>
+
+    <p>
+       All active users, in good standing with pledged amounts greater than zero:
```
MonkeyDo, line 6: "users" here might be misleading. Maybe "supporters" as in the title above, or "commercial supporters"?

mayhem: Fixed. Even though the author is the only person to ever see this report. :)

metabrainz/admin/views_test.py
```diff
@@ -18,3 +18,6 @@ def test_statsview_index(self):
 
     def test_statsview_top_ips(self):
         self.assertStatus(self.client.get(url_for('statsview.top_ips')), 302)
+
+    def test_statsview_supporters(self):
+        self.assertStatus(self.client.get(url_for('statsview.supporters')), 302)
```
paramsingh, line 23: why do we get a 302 here?

mayhem: I don't think the testing the admin pages has been fully setup before and given the limited audience of these pages, I can't find motivation to get the job done.

PR #349, opened 2021-01-19
Upgrade base image, thus python and therefor a bunch of dependencies.…

mayhem:
… Also fix on failing test.


requirements.txt
```diff
@@ -1,21 +1,23 @@
...
 Flask-Login==0.4.1
 Flask-SQLAlchemy==2.4.0
-Flask-Testing==0.7.1
-Flask-Uploads==0.2.1
-Flask-WTF==0.14.2
-psycopg2==2.7.3
-pytest-cov==2.5.1
-pytest==3.2.1
+Flask-Testing==0.8.1
+-e git://github.com/maxcountryman/flask-uploads.git#egg=Flask_Uploads
```
alastair, line 9: did you need to update this to a newer version because 0.2.1 doesn't support something? It'd be good to add an explicit tag or commit here to make sure that random changes to master don't break the build: 
`-e git://github.com/maxcountryman/flask-uploads.git@commit-hash#egg=Flask_Uploads`

alastair (approved):
If this runs for you then I'm happy with it, the changes look good.

PR #359, opened 2021-03-31
Upgrade brainzutils and sentry-sdk

amCap1712:
Apart from upgrading brainzutils and sentry-sdk, I have also removed configuring email logging. We do not have the required configurations anyways and never use it. I have also removed a loggers.py which does not seem to be used anywhere.


metabrainz/__init__.py
```diff
@@ -74,7 +74,6 @@ def create_app(debug=None, config_path = None):
 
     app.init_loggers(
         file_config=app.config.get('LOG_FILE'),
-        email_config=app.config.get('LOG_EMAIL'),
```
alastair, line 77: we should also remove this from the consul configs in docker-server-configs

metabrainz/model/payment.py
```diff
@@ -170,9 +170,7 @@ def process_paypal_ipn(cls, form):
         if form['payment_status'] != 'Completed':
             # TODO(roman): Convert to regular `logging.info` call when such detailed logs
             # are no longer necessary to capture.
-            get_sentry_client().captureMessage("PayPal: Payment is not completed",
-                                               level=logging.INFO,
-                                               extra={"ipn_content": form})
+            capture_message("PayPal: Payment is not completed", level="info", extra={"ipn_content": form})
```
alastair, line 173: just confirming, have you checked that this message gets reported properly?

amCap1712: Yes.

alastair (approved):
Based on our changes yesterday this looks fine to me, though I didn't test it myself.

PR #363, opened 2021-06-02
Add actions workflows

amCap1712:
Adding action workflows for building production images, running tests and creating release notes. We are now moving to a date based tag and release as in other *Brainz. We should also remove the Jenkins job after merging this.


mayhem (approved):
I dont know much about this stuff... but given that this isn't mission critical if it breaks, I'd say go for it.

PR #364, opened 2021-06-07
Simplify local development setup

amCap1712:
Simplify local development setup by removing unused and redundant files.
Use common setup for development and testing.
Make setup similar to LB.
Also, removed some outdated consul configs and scripts.


mayhem (approved):
I would love it if you could add a comment to the README that states how to make a release happen, now that push.sh is gone. After that, feel free to merge.

amCap1712:
There were actually two `push.sh`, one in root of the repo which I removed in this PR. Other is `docker/push.sh`, which is the same as the push.sh scripts in LB and other *Brainz projects.

mayhem:
Oh, I see. Still the comment about how to do a release by doing it at GH would be very good.

amCap1712:
Updated Readme.

PR #367, opened 2021-08-11
Validate integer query params

amCap1712:
and use the default value instead if param is not a valid integer

A bunch of errors related to this showed up in sentry, hence fixing it.

### DIFF
--- a/metabrainz/utils.py
+++ b/metabrainz/utils.py
@@ -4,6 +4,8 @@ import random
 import string
 import subprocess
 
+from flask import request
+
 
 def reformat_datetime(value, format='%x %X %Z'):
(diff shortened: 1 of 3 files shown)


metabrainz/utils.py
```diff
@@ -29,3 +31,10 @@ def build_url(base, additional_params=None):
         (url.scheme, url.netloc, url.path, url.params,
          urlencode(query_params), url.fragment)
     )
+
+
+def get_int_query_param(key, default):
```
alastair, line 36: docstring

alastair (approved):
I didn't run it, but this looks good!

PR #373, opened 2021-10-26
Simplify stripe keys configuration

amCap1712:
We don't need a STRIPE_KEYS and another STRIPE_TEST_KEYS. Just keep one and put the keys you are using there. For consul config this means, that test.meb config contains the test keys only while the prod one uses live keys only.


amCap1712:
Remember that deploy this and the consul config together.

mayhem (approved):
Much better!

PR #375, opened 2021-10-30
Subscriptions using Stripe

amCap1712:
The current implementation is able to create subscriptions but not cancel them. Cancelling stripe subscriptions requires us to store the donor's stripe customer id. This means that we'll need auth for stripe recurring subscriptions. Otherwise, we cannot be sure that a user's subscription is not messed up with by someone else. Before we go ahead to implement this, we should consider if this added complexoty is worth it.

### DIFF
--- a/metabrainz/model/payment.py
+++ b/metabrainz/model/payment.py
@@ -300,7 +300,23 @@ class Payment(db.Model):
         return options
 
     @classmethod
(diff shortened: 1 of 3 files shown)


mayhem:
Bump!

amCap1712:
We decided on IRC some months ago that manual cancellation for now is fine.

PR #377, opened 2022-01-12
Pin WTForms version

amCap1712:
WTForms 3.* moves arounds some fields so imports need to be updated, possible Flask-WTF may need to be updated as well to be compatible with it. For now, pin the latest 2.* version.


mayhem (approved):
Thanks!

PR #387, opened 2022-07-12
Support fetching replication v2 packets

mwiencek:
Since the last schema change, MB now generates dbmirror2 replication packets, which exist in the same volume and have a '-v2' suffix after the packet number.

This commit adds initial support for fetching such packets to help me test them with replication and JSON dumps.

For now I just made it so that you fetch them how they're named on disk (with '-v2' suffix) but we could alternatively add an API subpath like /api/musicbrainz/v2/ if that's desirable.

### DIFF
--- a/metabrainz/api/views/musicbrainz.py
+++ b/metabrainz/api/views/musicbrainz.py
@@ -96,12 +96,12 @@ def replication_info():
     })
 
 
-@api_musicbrainz_bp.route('/replication-<int:packet_number>.tar.bz2')
-@token_required
-@tracked
-def replication_hourly(packet_number):
+def _replication_hourly(packet_number, v2):
     directory = current_app.config['REPLICATION_PACKETS_DIR']
-    filename = 'replication-%s.tar.bz2' % packet_number
+    if v2:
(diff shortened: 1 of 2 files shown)


mwiencek:
> One mistake I've regretted for almost 20 years is not having broken the replication packets into subdirs so that we don't overload one single directory. I wonder if we should improve that for v2 packets?

Yes!  I think we could fix them for the old packets too (it's still annoying accidentally hitting `ls<tab>` in there).

Having the subdirs derived from the packet number would let us keep the API the same.  The only alternative I can think of is by date, but you can't reliably determine the date given a packet number.

So maybe something like -

/v1/1/5/1/0/replication-151097.tar.bz2
/v2/1/5/1/0/replication-151097-v2.tar.bz2

Then we can (1) create hard links under the new directories that point to the original files, (2) update the metabrainz container to read from the new directories, and (3) remove the original files.

mwiencek:
Three levels sounds good.  I'd rather a separate PR for that since I'll have to update MBS to make the correct directory structure first.

mayhem (approved):
Thanks!

amCap1712 (approved):
LGTM, Thanks!

PR #390, opened 2022-09-09
Send invoices

mayhem:
This PR implements the creation of sending invoices. We shouldn't need this, but QB's invoice sending servers have been blacklisted on spam lists, so many of our invoices were not reaching their destination. Grrr.

This PR does the following:
- After logging into the create invoices page, the credientials for the QB API are stored in the MeB redis.
- Using manage.py in the running metabrainz-prod container, invoices can be sent from our mail server.
- The script fires up, examines the latest 300 invoices, marks voided invoices as sent (so they are ignored in the future). Any invoices that are unclear will prompt the user for action.
- Any invoices that are sent are marked as sent and accounting@ is cc'ed on the mail.
- The bodies of the emails being sent look very similar to the emails that are going out now.

### DIFF
--- a/metabrainz/invoices/send_invoices.py
+++ b/metabrainz/invoices/send_invoices.py
@@ -16,6 +16,7 @@ from intuitlib.exceptions import AuthClientError
 from brainzutils import cache
 from metabrainz.mail import send_mail
 
+SEND_DELAY = 5
 MAIL_BODY = """To: %s %s (%s)
 
 Here is your next invoice! For information on how to pay us, please see https://metabrainz.org/payment
@@ -78,6 +79,7 @@ class QuickBooksInvoiceSender():
             return False
 
     def send_invoice(self, client, invoice, customer):
+        print("  sending invoice")
         emails = [ e.strip() for e in str(invoice.BillEmail).split(",") ]
         emails.append("accounting@metabrainz.org")
         text = MAIL_BODY % (customer.GivenName,
@@ -98,6 +100,9 @@ class QuickBooksInvoiceSender():
             from_addr="accounting@metabrainz.org",
             from_name="MetaBrainz Accounting Department"
         )
+        self.mark_invoice_sent(client, invoice)
+        print("  wait")
+        time.sleep(SEND_DELAY)
 
     def send_invoices(self):
 
--- a/metabrainz/invoices/send_invoices.py
+++ b/metabrainz/invoices/send_invoices.py
@@ -1,7 +1,9 @@
 import datetime 
+import io
 import time
 from decimal import Decimal
 
+from flask import current_app
 from intuitlib.enums import Scopes
 from intuitlib.client import AuthClient
 from quickbooks import QuickBooks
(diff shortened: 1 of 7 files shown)


metabrainz/invoices/send_invoices.py
```diff
@@ -0,0 +1,165 @@
...
+            if invoice.EmailStatus == "NotSet":
+                print("  To '%s' marked as NotSet." % customer.DisplayName)
+                while True:
+                    print("  Send [s], Mark sent [m], Ignore [i]:", end="")
+                    resp = input().strip().lower()
+                    if resp is None or len(resp) == 0 or resp[0] not in "smi":
+                        print("  select one of the given options!")
+
+                    if resp[0] == "s":
+                        self.send_invoice(client, invoice)
```
amCap1712, line 152: missing `customer` parameter at end.

metabrainz/invoices/send_invoices.py
```diff
@@ -0,0 +1,165 @@
...
+            invoice.save(qb=client)
+            return True
+        except exceptions.ValidationException as err:
+            print(err.detail)
+            return False
+
+    def send_invoice(self, client, invoice, customer):
+        """ Given a client, invoice and its customer object, send the invoice to the customer. """
+
+        print("  sending invoice")
```
amCap1712, line 89: Here and elsewhere in the file, instead of `print` can use `current_app.logger.info`

metabrainz/invoices/send_invoices.py
```diff
@@ -0,0 +1,165 @@
...
+            customer = Customer.get(int(invoice.CustomerRef.value), qb=client)
+            if customer.Notes.find("donotsend") >= 0:
+                print("  marking donotsend invoice %s as sent, without sending." % invoice.DocNumber)
+                self.mark_invoice_sent(client, invoice)
+
+            if invoice.EmailStatus == "NotSet":
+                print("  To '%s' marked as NotSet." % customer.DisplayName)
+                while True:
+                    print("  Send [s], Mark sent [m], Ignore [i]:", end="")
+                    resp = input().strip().lower()
```
amCap1712, line 147:
```suggestion
                    resp = input("  Send [s], Mark sent [m], Ignore [i]:").strip().lower()
```

metabrainz/invoices/send_invoices.py
```diff
@@ -0,0 +1,165 @@
...
+                    if resp is None or len(resp) == 0 or resp[0] not in "smi":
+                        print("  select one of the given options!")
+
+                    if resp[0] == "s":
+                        self.send_invoice(client, invoice)
+                        print("  invoice sent!")
+                        break
```
amCap1712, line 154: since breaking in all cases, break can be done unconditionally after the if.

metabrainz/invoices/send_invoices.py
```diff
@@ -0,0 +1,165 @@
...
+                    elif resp[0] == "m":
+                        self.mark_invoice_sent(client, invoice)
+                        print("  invoice marked as sent, without being sent!")
+                        break
+                    else:
+                        break
+
+                continue
```
amCap1712, line 162: redundant

amCap1712 (approved):
I do not know much about QuickBooks so can only give feedback on general code and nits. Feel free to ignore.

Since it works overall in production, then LGTM. Thanks!

PR #392, opened 2022-10-14
Minor fixes (and fix a long standing bug) for the send invoices feature.

mayhem:
These two minor fixes fix a problem where copied invoices contained incorrect DeliveryInfo (which is now reset to default for each new invoice) and where a do no send invoice was actually sent, because of a missing continue statement.


amCap1712 (approved):
LGTM, Thanks!

PR #393, opened 2022-10-18
Send invoice reminders

mayhem:
This part was still missing, but now we can send invoice reminders by just running one command. Much easier!

Much like sending invoices, but a slightly different workflow.


amCap1712 (approved):
LGTM, Thanks!

PR #407, opened 2023-05-04
Improved datasets pages

mayhem:
This PR explodes the simple datasets page into a series of pages that explain out datasets and feed users into a path of signing up (optional) with MetaBrainz and then leads the users to the data download page. This page will not be indexed by Google, so that people can't just skip to the end and download our data. Clearly there will be many ways that a user can circumvent this sign-up process, so it is not meant to be fool proof, but rather meant for us to collect more data about the data use of our users and to make it clear that people need to support us.

All of the text and design changes have already been reviewed multiple times, so this PR should mostly check the technical aspects of this PR for expediency.


amCap1712 (approved):
LGTM, Thanks!

PR #409, opened 2023-05-05
Add dotsrc

mayhem:
Adding the DK dotsrc mirror and changing other links to https.


mayhem:
Available for preview here: https://test.metabrainz.org/datasets/download

reosarevok (approved):
The idea seems fine, I did not copy and paste every link to test though (can do that if you put it on test)

PR #411, opened 2023-05-05
Improve text according to aerozols's improvements

mayhem:
As discussed in IRC.


reosarevok (approved):
I think this is good now :)

PR #412, opened 2023-05-15
Remove the reports link

mayhem:
Since the reports have not been updated since 2017, lets remove the link.

### DIFF
--- a/metabrainz/templates/navbar.html
+++ b/metabrainz/templates/navbar.html
@@ -44,7 +44,6 @@
         <li class="dropdown">
           <a class="dropdown-toggle" data-toggle="dropdown" href="#">{{ _('Reports ') }}<span class="caret"></span></a>
           <ul class="dropdown-menu" role="menu">
-            <li><a href="{{ url_for('annual_reports.index') }}">{{ _('Annual Reports') }}</a></li>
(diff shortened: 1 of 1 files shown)


yvanzo:
It seems to be paradoxical as during the latest summits we always talked about giving more visibility to our activities. Personally these reports were very informative and reassuring before I joined the project. Where can people get similar information from nowadays?

* Would displaying these as archived instead of removing be acceptable if a patch popped up?
* Would resuming reports in a more lightweight form, would be useful to achieve the same transparency and visibility to MetaBrainz activities? IIUC, the problem was that it was taking too much time to the same person. I can volunteer.

mayhem:
I'm ok with the idea of calling them historical in some way.

But I stopped doing the reports because many of our metrics are no longer capturable -- we dont have traffic stats, for instance. Furthermore, it took FIVE years for anyone to notice that these were no longer being updated.

To me this was a clear indication that I was wasting my time on writing reports (2-3 days usually), so I stopped doing.

If we want to resurrect these, then it would need to be a community effort led by someone other than me.

reosarevok (approved):
I agree with keeping the page and clarifying it, but I also agree with removing it from the main menu since it's not maintained. It could be linked from somewhere else, like https://metabrainz.org/about ?

PR #414, opened 2023-05-22
Archive annual reports

yvanzo:
It follows the comments to the pull request #412 (to replace it).

- Add archive notice to annual reports
- Reword the menu item “Annual Reports” (and same strings) as “Archived Annual Reports”

### DIFF
--- a/metabrainz/reports/annual_reports/files/2018/content.html
+++ b/metabrainz/reports/annual_reports/files/2018/content.html
@@ -1,0 +1,29 @@
+<h2>Archive notice</h2>
+
+<p>
+Annual reports from 2006 to 2017 are still available for history.
+There was no annual report in this form before 2006.
+Since 2018, no annual report has been produced as it was too much overhead for one person.
+Most of the information it was gathering is nowadays available as follows:
+</p>
+
+<ul>
+  <li>
+    <b>Year in review:</b>
+    Most major announcements are published on our
+    <a href="https://blog.metabrainz.org/">blog</a>.
+    Some minor annoucements are made during our
+    <a href="https://community.metabrainz.org/tag/metabrainz-meeting-notes">weekly meetings</a>.
+  </li>
+  <li>
+    <b>Finances:</b>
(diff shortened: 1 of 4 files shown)


metabrainz/reports/annual_reports/files/2018/content.html
```diff
@@ -0,0 +1,29 @@
+<h2>Archive notice</h2>
+
+<p>
+Annual reports from 2006 to 2017 are still available for history.
```
reosarevok, line 4:
```suggestion
Annual reports from 2006 to 2017 are still available as historical data.
```
Maybe?

metabrainz/reports/annual_reports/files/2018/content.html
```diff
@@ -0,0 +1,29 @@
...
+There was no annual report in this form before 2006.
+Since 2018, no annual report has been produced as it was too much overhead for one person.
```
reosarevok, line 6: I'm not sure what would be good here, but this seems like giving excuses rather than saying we just don't do this anymore.

metabrainz/reports/annual_reports/files/2018/content.html
yvanzo: I’m missing the difference but I’m not native English speaker. Is “history” too much here? Would “for the record” be more appropriate?
```suggestion
Annual reports from 2006 to 2017 are still available for the record.
```

metabrainz/reports/annual_reports/files/2018/content.html
yvanzo: If you’re not sure, I’m pretty sure that’s good. As we discussed about it with @mayhem it appeared to be the main reason these reports have been indefinitely suspended. Explaining why there is less transparency on our activity at the moment seems to be the bare minimum.

yvanzo:
```suggestion
Since 2018, no annual report has been produced as there wasn’t enough interest into it.
```

PR #416, opened 2023-05-25
Remove user to supporter throughout the codebase - 2

amCap1712:
Rename the actual tables and some changes that I missed the last time around.


admin/sql/create_foreign_keys.sql
```diff
@@ -2,21 +2,21 @@ BEGIN;
 
 ALTER TABLE token
   ADD CONSTRAINT token_owner_id_fkey FOREIGN KEY (owner_id)
-  REFERENCES "user" (id) MATCH SIMPLE
+  REFERENCES "supporter" (id) MATCH SIMPLE
```
mayhem, line 5: we could finally get rid of the " " since user is not a reserved name, right?

mayhem (approved):
One minor cleanup possible, but thats it. Thanks!

admin/sql/create_foreign_keys.sql
amCap1712: yup. done

PR #432, opened 2023-09-05
Amend pull request #413: Revert template path

yvanzo:
# Fix regression

It fixes the following error that was in server logs:

    jinja2.exceptions.TemplateNotFound: admin/stats/users.html

# Draft progress

* [x] Deploy to test instance
* [x] Test admin interface and check server logs


mayhem (approved):

yvanzo:
Tested visiting all admin pages and nothing showed up in server logs.

PR #434, opened 2023-09-13
Remove alastair

mayhem:
Remove alastair from the team list. :(


MonkeyDo (approved):
I am quite sad to see Alastair removed from the team :(

PR #439, opened 2023-10-10
enable_global no longer exists

mayhem:
Not sure why this died now, but sending invoices no longer worked. Removing it and everything works. Odd.


MonkeyDo (approved):
enable_global was deprecated in 2021 and removed in March of 2023:
https://github.com/ej2/python-quickbooks/releases/tag/0.9.3

https://github.com/ej2/python-quickbooks/blob/master/CHANGELOG.rst

I know nothing of QuickBooks, but it looks sane to remove it.

PR #447, opened 2023-11-15
Drop mirror links

reosarevok:
We no longer use mirrors, so the first commit drops them from the datasets page (which is certainly wanted).

AFAICT this is the only support these two supporters were providing, so the second commit drops them from the sponsors page and the footer - let me know if that's not correct and I can drop that commit.

Untested but seems trivial.


mayhem (approved):
thanks!

PR #448, opened 2023-11-20
Clarify Privacy Policy about web API

yvanzo:
Two small updates to our [Privacy Policy](https://metabrainz.org/privacy):

1. It explicitly mentions web API calls in the paragraph about server logs. It comes after a [user’s question on IRC](https://chatlogs.metabrainz.org/libera/metabrainz/msg/5250667/).
2. <del>It drops FTP as we don’t use it anymore, datasets now have to be downloaded from web servers through HTTPS even though `ftp` remains in a couple of subdomain names for now.</del>

### DIFF
--- a/metabrainz/templates/index/privacy.html
+++ b/metabrainz/templates/index/privacy.html
@@ -54,8 +54,8 @@
   <h3>{{ _('Web and FTP access logs') }}</h3>
 
   <p>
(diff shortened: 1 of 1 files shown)


yvanzo:
I reverted dropping FTP as it is still up and running even though not listed from our download pages.

zas (approved):
LGTM

PR #449, opened 2023-11-24
Meb 143: Update projects page

Aerozol:
MEB-143: Update projects page

Edit https://metabrainz.org/projects to remove AcousticBrainz and MessyBrainz.
Remove defunct assets/logos, only used by AcousticBrainz and MessyBrainz (2 from frontend/img/projects, 1 from  frontend/img/logos).

Searched the org repository for other uses for these logos/file paths and didn't find any, so assume it's safe to remove them, but I would appreciate someone else double checking.


yvanzo (commented):
The only mention left of AcousticBrainz is in the GDPR, which is fine since the AB website and API continue to be available for now even though it stopped collecting data. Would it be appropriate to have a glorious history page/section for past projects?

Aerozol:
Not a bad idea yvanzo, I've done that in https://github.com/metabrainz/metabrainz.org/pull/452
We can compare and see which is preferred :)

PR #452, opened 2023-11-30
MEB-143: Update projects page v2

Aerozol:
Remove AcousticBrainz and MessyBrainz from:
https://metabrainz.org/projects

This pull moves them to a 'Past Projects' section instead of removing them, as suggested by yvanzo in: https://github.com/metabrainz/metabrainz.org/pull/449


mayhem (approved):
Thank you!

PR #453, opened 2024-01-01
Update conflict-policy.html

greydoubt:
cleaned up for readability, i had trouble understanding the difference versus private / public at first


yvanzo (approved):
Thank you for the phrasing improvements.

PR #461, opened 2024-04-05
Fix invoice numbers

mayhem:
Quickbooks changed something on how invoice numbers are created... and broke something very simple. This PR makes the invoice number assignment automatic now based on the highest invoice number found.


amCap1712 (approved):
LGTM, Thanks!

PR #470, opened 2024-07-16
Support EUR payments in stripe

amCap1712:
The EUR account is different from the USD account and hence using a different set of API keys.

### DIFF
--- a/metabrainz/payments/stripe/views.py
+++ b/metabrainz/payments/stripe/views.py
@@ -23,6 +23,11 @@ def pay():
         return redirect(url_for("payments.error", is_donation=is_donation))
 
     is_recurring = form.recurring.data
+    currency = form.currency.data
+    if currency.lower() == "usd":
+        api_key = current_app.config["STRIPE_KEYS"]["USD"]["SECRET"]
+    else:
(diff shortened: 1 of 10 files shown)


metabrainz/payments/stripe/views.py
```diff
@@ -77,18 +82,24 @@ def pay():
...
+        session = stripe.checkout.Session.create(**session_config, api_key=api_key)
         return redirect(session.url, code=303)
     except Exception as e:
         current_app.logger.error(e, exc_info=True)
         return redirect(url_for("payments.error", is_donation=is_donation))
 
 
-@payments_stripe_bp.route("/webhook/", methods=["POST"])
-def webhook():
+@payments_stripe_bp.route("/webhook/<currency>/", methods=["POST"])
```
mayhem, line 92: We'll need to install a new webhook and change the other one at deploy time?

mayhem (approved):
Looking good so far - I've pushed one small change to the payments page to improve the working, should be good to go. Thanks!

metabrainz/payments/stripe/views.py
amCap1712: Change it at deploy time.

PR #477, opened 2024-08-30
MeB OAuth - latest iteration

amCap1712:


MonkeyDo:
I have applied the last bits of feedback to the OAuth applications pages based on the comments of the design figma
https://www.figma.com/design/L0qNv3z5vSkNDoiX7BzaY5/MetaBrainz-redesign?node-id=0-1&node-type=canvas

We might eventually want to restyle these pages, but as aerozol commented on that figma: "Looks good, just needs to be functional"

frontend/js/src/forms/DeleteApplication.tsx
```diff
@@ -0,0 +1,77 @@
...
+                <a href={cancel_url} className="btn btn-default">
+                  Cancel
+                </a>
+              </div>
+              <div className="col-md-1" style={{ marginLeft: "8px" }}>
+                <button type="submit" className="btn btn-danger">
+                  Delete
+                </button>
+              </div>
+            </div>
```
MonkeyDo, line 57: On smaller sizes the button stack up weirdly.
image

You can use a bootstrap btn-group instead of using col-* utilities like so:

```suggestion
            <div className="btn-group">
              <a href={cancel_url} className="btn btn-default">
                Cancel
              </a>
              <button type="submit" className="btn btn-danger" style={{ marginLeft: "8px" }}>
                Delete
              </button>
            </div>
```



Result 
image

PR #482, opened 2024-09-11
Fetch musicbrainz_row_id from MB for non-supporters

amCap1712:
A supporter account is not required for donations, hence fallback to checking MB db for musicbrainz_row_id if the specified editor_name is not associated with a supporter.

### DIFF
--- a/config.py.example
+++ b/config.py.example
@@ -7,6 +7,7 @@ SECRET_KEY = "CHANGE_THIS"
 
 # DATABASE
 SQLALCHEMY_DATABASE_URI = "postgresql://metabrainz:metabrainz@db:5432/metabrainz"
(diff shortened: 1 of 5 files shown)


config.py.example
```diff
@@ -7,6 +7,7 @@ SECRET_KEY = "CHANGE_THIS"
 
 # DATABASE
 SQLALCHEMY_DATABASE_URI = "postgresql://metabrainz:metabrainz@db:5432/metabrainz"
+SQLALCHEMY_DATABASE_URI = ""
```
mayhem, line 10: This is odd -- why are you doing this?

amCap1712: uhh, my bad. typo

PR #483, opened 2024-09-11
Prefill donation form if user is logged in

amCap1712:


mayhem (approved):

MonkeyDo (commented):
@mayhem also mentioned that we might want to allow pre-filling the donation amount as well, I reckon with a url parameter (what about currency?).

I'm not sure if that was specifically and only for the LB donations page, but it seems like a logical thing to have available.

amCap1712:
@MonkeyDo Added prefill for donation and currency. Also, added prefill for editor name from url with a fallback to current logged in user's name if editor is not supplied in the url.

PR #485, opened 2024-10-06
Replacing "IRC" with "Chat" or "ChatBrainz"

Aerozol:
As mentioned by mayhem at the summit, updated "IRC" references to "Chat" or "ChatBrainz" references and changed the relevant links.

Will need testing, sorry, but hopefully this is most of the work done.

### DIFF
--- a/metabrainz/messages.pot
+++ b/metabrainz/messages.pot
@@ -1,62 +1,54 @@
 # Translations template for PROJECT.
-# Copyright (C) 2017 ORGANIZATION
+# Copyright (C) 2024 ORGANIZATION
 # This file is distributed under the same license as the PROJECT project.
-# FIRST AUTHOR <EMAIL@ADDRESS>, 2017.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2024.
 #
 #, fuzzy
 msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2017-01-27 14:44+0100\n"
+"POT-Creation-Date: 2024-10-29 09:23+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Generated-By: Babel 2.3.4\n"
+"Generated-By: Babel 2.16.0\n"
 
-#: metabrainz/payments/forms.py:9
-msgid "You need to specify amount!"
-msgstr ""
-
-#: metabrainz/payments/forms.py:21
-msgid "You need to specify invoice number!"
-msgstr ""
-
-#: metabrainz/payments/views.py:105
+#: metabrainz/payments/views.py:155
 msgid ""
 "Thank you for making a donation to the MetaBrainz Foundation. Your "
 "support is greatly appreciated! It may take some time before your "
 "donation appears in the list due to processing delays."
 msgstr ""
 
-#: metabrainz/payments/views.py:112
+#: metabrainz/payments/views.py:162
 msgid ""
 "Thank you for making a payment to the MetaBrainz Foundation. Your support"
 " is greatly appreciated!"
 msgstr ""
 
-#: metabrainz/payments/views.py:123
+#: metabrainz/payments/views.py:173
 msgid ""
 "We're sorry to see that you won't be donating today. We hope that you'll "
 "change your mind!"
 msgstr ""
 
-#: metabrainz/payments/views.py:129
+#: metabrainz/payments/views.py:179
 msgid ""
 "We're sorry to see that you won't be paying today. We hope that you'll "
 "change your mind!"
 msgstr ""
 
-#: metabrainz/payments/views.py:143
+#: metabrainz/payments/views.py:193
 msgid ""
 "We're sorry, but it appears we've run into an error and can't process "
 "your donation."
 msgstr ""
 
-#: metabrainz/payments/views.py:149
+#: metabrainz/payments/views.py:199
 msgid ""
 "We're sorry, but it appears we've run into an error and can't process "
 "your payment."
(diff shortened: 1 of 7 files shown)


mayhem:
Hey, thanks for tackling this. I'll pay attention to it once the back-to-aback summits have stopped their toll on me.

Aerozol:
No rush, take it easy!

metabrainz/templates/index/contact.html
```diff
@@ -87,22 +87,31 @@ <h3>{{ _('Forums') }}</h3>
...
-    instead. Please feel free to browse the
-    <a href="%(chatlogs_url)s">IRC log archives</a>.',
+    {{ _('The MetaBrainz/MusicBrainz/ListenBrainz team and community hang out
+    on our Matrix, IRC and Discord channels. Visit the
+    <a href="https://musicbrainz.org/doc/Communication/ChatBrainz">ChatBrainz</a>
+    page and join the chat. You can also browse the
+    <a href="%(chatlogs_url)s">chat log archives</a>.',
     irc_url_mb='ircs://irc.libera.chat:6697/musicbrainz',
     irc_url_meb='ircs://irc.libera.chat:6697/metabrainz',
     irc_webchat_url='https://kiwiirc.com/nextclient/irc.libera.chat/#musicbrainz,#metabrainz',
```
reosarevok, line 99: These URLs should probably be removed since they're no longer used. And the ChatBrainz one should probably be used in a similar way as these were rather than just pasting it as-is inside the string.

metabrainz/templates/index/contact.html
```diff
@@ -87,22 +87,31 @@ <h3>{{ _('Forums') }}</h3>
...
     chatlogs_url='http://chatlogs.metabrainz.org/') }}
   </p>
 
+  <h3>{{ _('Social Media') }}</h3>
+  <p>
+    {{ _('Nobody\'s perfect. You can follow us on these social channels:') }}
```
reosarevok, line 105: Maybe use a proper apostrophe and then you don't need to escape it ;)  Although I'm not sure why the "Nobody's perfect" bit is there anyway? :sweat_smile:

metabrainz/templates/index/contact.html
```diff
@@ -87,22 +87,31 @@ <h3>{{ _('Forums') }}</h3>
...
+    <a href="https://musicbrainz.org/doc/Communication/ChatBrainz">ChatBrainz</a>
+    page and join the chat. You can also browse the
+    <a href="%(chatlogs_url)s">chat log archives</a>.',
     irc_url_mb='ircs://irc.libera.chat:6697/musicbrainz',
     irc_url_meb='ircs://irc.libera.chat:6697/metabrainz',
     irc_webchat_url='https://kiwiirc.com/nextclient/irc.libera.chat/#musicbrainz,#metabrainz',
     chatlogs_url='http://chatlogs.metabrainz.org/') }}
   </p>
 
+  <h3>{{ _('Social Media') }}</h3>
```
reosarevok, line 103: We use sentence case for other headers.

metabrainz/templates/index/contact.html
```diff
@@ -87,22 +87,31 @@ <h3>{{ _('Forums') }}</h3>
...
+  <p>
+    {{ _('Nobody\'s perfect. You can follow us on these social channels:') }}
+     <ul>
+      <li>{{ _('<a href="https://blog.metabrainz.org/">Blog</a>') }}</li>
+      <li>{{ _('<a href="https://x.com/MetaBrainz">Twitter/X</a>') }}</li>
+      <li>{{ _('<a href="https://mastodon.social/@metabrainz">Mastodon</a>') }}</li>
+      <li>{{ _('<a href="https://bsky.app/profile/metabrainz.org">Bluesky</a>') }}</li>
+      <li>{{ _('<a href="https://www.instagram.com/MetaBrainz/">Instagram</a>') }}</li>
```
reosarevok, line 111: Ideally these should also be done with variables such as `blog_url`, `mastodon_url`, etc - that way if the URL changes later (say, because some idiot buys Bluesky and renames it Y) it won't change the actual string translators will see.

Alternatively (and possibly even better?) you could just take the URL bit out of the translatable bit and leave just the title, since that seems to be what we already do in the footer?

metabrainz/messages.pot
```diff
@@ -1069,7 +1065,7 @@ msgstr ""
 msgid ""
 "Hired in the fall of 2015 as the MetaBrainz Community Manager, Freso is "
 "in charge of most of the public communication\n"
-"          channels (IRC, forums, blog, etc.) and various community "
+"          channels (Chat, forums, blog, etc.) and various community "
```
reosarevok, line 1068: Did you do the `messages.pot` changes by hand? I assume the `.pot` file is automatically generated by some script and we should run that instead.

reosarevok (changes requested):

metabrainz/templates/index/contact.html
Aerozol: Thanks for giving feedback! I was hoping someone else would come along and do everything 'proper' but it turns out this wasn't hard to do. I've made the suggested change (hopefully correctly)

metabrainz/templates/index/contact.html
Aerozol: We can use proper apostrophes? Cool! I guess we didn't use them wherever I copied this from, but I'll do it next time.
I've removed the "Nobody's perfect" bit anyway, I was just feeling down on "social media" at the time, but nobody likes a downer

metabrainz/templates/index/contact.html
Aerozol: Fixed, thanks

metabrainz/messages.pot
Aerozol: This is the only bit I do not know how to do. Happy to tackle it with guidance, otherwise I will leave it to the devs

metabrainz/templates/index/contact.html
reosarevok: You had left the quotes in the wrong place, but I fixed that now :)

metabrainz/messages.pot
reosarevok: The info was on the readme file, but it was actually semi-incorrect and I had to play with it until I got it to work. Amended the readme in a separate commit. Also, it seems nobody had updated this since... 2017 or so? (which is why the string you were modifying here was still about Freso being in charge of communication).

reosarevok:
I think this is now ready but one of @mayhem or @amCap1712 should take a quick look at least.

mayhem (approved):
Yep, happy. Thanks! Will release once the funding.json stuff is done.

PR #489, opened 2024-11-14
Change nag to $5 for one month

mayhem:
Not much else to say.


MonkeyDo (approved):
Is there a process for when we want to change this value?
Does the board need to be involved or informed?

I see that the feature was added and last touched in 2014, so I guess not :)

PR #495, opened 2024-12-21
Support user login and registration

amCap1712:
TBA


admin/sql/create_indexes.sql
```diff
@@ -2,4 +2,6 @@ BEGIN;
 
 -- TODO: Add some, if needed.
```
mayhem: cleanup?

metabrainz/index/forms.py
```diff
@@ -0,0 +1,75 @@
...
+
+from flask_wtf import FlaskForm
+from wtforms import validators
+from wtforms.fields import SelectMultipleField, EmailField, StringField
+from wtforms.validators import DataRequired
+from wtforms.widgets.core import ListWidget, CheckboxInput
+
+
+class MeBFlaskForm(FlaskForm):
+
```
mayhem: docstrings.

metabrainz/user/email.py
```diff
@@ -0,0 +1,68 @@
+import hashlib
+from datetime import datetime
+
+from flask import url_for, request, render_template, current_app
+
+from brainzutils.mail import send_mail
+from metabrainz.model.user import User
+
+VERIFY_EMAIL = "verify-email"
```
mayhem: These are oddly named. What do they do?

oauth/views.py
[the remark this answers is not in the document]
julian45: Also, since `Cache-Control: no-store` per the preceding line, I believe this setting is also kind of ignored since `no-store` seems to be the most restrictive directive among possible `Cache-Control` values; see MDN [Pragma](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma#directives), [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#use_cases) pages.

PR #498, opened 2025-01-15
Increase font size

mayhem:
As per our discussion, upping font size to 16px, save for the navbar. for reasons.


MonkeyDo (approved):
:rocket:

PR #500, opened 2025-01-24
MEB-167: Clarify name in donations means real name 

reosarevok:
### Implement MEB-167

# Description
Apparently, even though a donor only enters their username from the donate page, their real name is also published as per their billing info unless the donation is anonymous. While "your name" is indicated under the anonymous checkbox, given there's nothing indicating your real name will be used at all, this is only easily understandable as "your username" which you are providing. Just to make sure, I asked my wife what she thought would get published by checking the form and she just said "well, that username there, right?".

It's even more confusing given just under the form we say "The personal information provided to the MetaBrainz Foundation during the donation process will not be shared with anyone" - most people, without more information being provided, would I think assume that includes their real name - I know I would have.

We had a person in support asking (very nicely) for us to remove their real name. I feel they could have been a lot less nice given the surprise of seeing their name effectively doxxed with little warning.

If we are going to publish real names (I assume this is some US requirement to prove real humans are donating or something) then we need to be *very clear* that this will happen. I'm not sure this change is enough, but at least it's not entirely unspecified anymore.

# Testing
Loaded the page locally to make sure it shows up fine.

### DIFF
--- a/metabrainz/messages.pot
+++ b/metabrainz/messages.pot
@@ -1,14 +1,14 @@
 # Translations template for PROJECT.
-# Copyright (C) 2024 ORGANIZATION
+# Copyright (C) 2025 ORGANIZATION
 # This file is distributed under the same license as the PROJECT project.
-# FIRST AUTHOR <EMAIL@ADDRESS>, 2024.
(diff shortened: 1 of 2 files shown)


mayhem (approved):
Thanks!

PR #507, opened 2025-05-28
Add fetch, mark and delete endpoints for notifications

fettuccinae:
Added
1. Fetch endpoint (/notification/fetch)  to get notifications with specified parameters.
2. Mark endpoint (/notification/mark-read)  to mark notifications as read or unread.
3. Delete endpoint (/notifcation/delete) to delete notifications.
4. Tests for the above endpoints and db functions.

In dev environment, `load_user_from_db` returns a hardcoded test user.


metabrainz/api/decorators.py
```diff
@@ -10,9 +10,18 @@ def token_required(f):
     def decorated(*args, **kwargs):
         access_token = request.args.get('token')
         if not access_token:
-            return Response("You need to provide an access token!\n", status=400)
-        if not Token.is_valid(access_token):
+            access_token = request.headers.get('Authorization')
```
fettuccinae, line 13: Should we allow users to validate their token through both a query parameter and the Authorization header?

metabrainz/notifications/views.py
```diff
@@ -0,0 +1,228 @@
...
+    if until_ts is None:
+        until_ts = datetime.now(timezone.utc)
+    else:
+        try:
+            until_ts = datetime.fromtimestamp(float(until_ts), tz=timezone.utc)
+        except ValueError:
+            return Response('Invalid Until_Timestamp ', 400)
+
+    project = request.args.get('project')
+    projects=[]
```
fettuccinae, line 74: The description for this parameter was a bit ambigous in the document. 

>  Filter which project to limit notifications for; if not specified, return for all projects.
>	project		Required. Must be a string representing the lowercase, no
space name of the project that is sending the notification.

So, i went with the first one, If not specified, return for all projects.

metabrainz/notifications/views.py
```diff
@@ -0,0 +1,228 @@
...
+
+    """
+    token = g._access_token
+    user_id = token.owner_id
+
+    count = request.args.get('count')
+    if count is None:
+        count = DEFAULT_NOTIFICATION_FETCH_COUNT
+    else:
+        count = int(count)
```
fettuccinae, line 56: Should we add some upper and lower bounds for the `count `and `offset ` parameters?

mayhem (changes requested):
Making good progress -- don't let the number of comments get you down -- you're doing really well!

metabrainz/notifications/views.py
```diff
@@ -0,0 +1,228 @@
...
+        offset = int(offset)
+
+    until_ts = request.args.get('until_ts')
+    if until_ts is None:
+        until_ts = datetime.now(timezone.utc)
+    else:
+        try:
+            until_ts = datetime.fromtimestamp(float(until_ts), tz=timezone.utc)
+        except ValueError:
+            return Response('Invalid Until_Timestamp ', 400)
```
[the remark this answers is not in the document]
fettuccinae: done

metabrainz/db/notification.py
```diff
@@ -0,0 +1,164 @@
...
+        """)
+        
+        result = connection.execute(query, params).fetchall()
+        return [
+            {"id": r.id, "user_id": r.musicbrainz_row_id, "project": r.project, "body": r.body,
+            "template_id": r.template_id,"sent_at": r.created, "read": r.read, "important": r.important} for r in result
+            ]
+
+
+def validate_notification_ids_for_user(user_id: int, notification_ids: Tuple[int, ...]) -> bool:
```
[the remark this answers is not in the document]
fettuccinae: It was to be used when an end user was sends a request to the endpoint. If they sent incorrect IDs, we could stop before the transaction started and not update anything.

metabrainz/db/notification.py
```diff
@@ -32,20 +36,20 @@ def fetch_notifications(user_id: int, projects: List[str], count: int, offset: i
...
+                        ,created AS sent_at
+                        ,read
+                        ,important
                 FROM
                         notification    
                 WHERE
                         musicbrainz_row_id = :user_id
-                        AND created <= :until_ts
-                        AND project in :projects
+                        AND (:until_ts IS NULL OR created <= :until_ts)
```
mayhem, line 51: Oh, I really like this solution, I'll have to remember that for my own work.

metabrainz/errors.py
```diff
@@ -22,3 +22,62 @@ def internal_server_error(error):
     @app.errorhandler(503)
     def service_unavailable(error):
         return render_template('errors/503.html', error=error), 503
+    
+    @app.errorhandler(APIError)
```
mayhem, line 26: Oh right -- thanks for copying those into this repo.

mayhem (changes requested):
Looking pretty good now -- lets now focus on the TODOs that are left and also to write some tests for the functions/endpoints that you've added.

metabrainz/db/notification.py
```diff
@@ -0,0 +1,121 @@
...
+                FROM
+                        notification    
+                WHERE
+                        musicbrainz_row_id = :user_id
+                        AND (:until_ts IS NULL OR created <= :until_ts)
+                        AND (:projects IS NULL OR project IN :projects)
+                        AND (:unread_only = FALSE OR read = FALSE)
+                ORDER BY
+                        created DESC
+                OFFSET
```
fettuccinae, line 56: I dont think `OFFSET `should come after `LIMIT ` because if we do it that way, we'll be skippin rows after limiting the result.

metabrainz/notifications/views.py
```diff
@@ -92,9 +92,10 @@ def get_notifications(user_id: int):
         unread_only=False
     elif unread_only == 't':
         unread_only=True
+    else:
+        raise APIBadRequest('Invalid unread_only option')
 
-    # data = fetch_notifications(user_id, tuple(projects), count, offset, until_ts, unread_only)
-    data = fetch_notifications(user_id=user_id)
+    data = fetch_notifications(user_id, tuple(projects), count, offset, until_ts, unread_only)
```
fettuccinae, line 98: Ooopsie here, I had forgotten to uncomment this line.

metabrainz/notifications/views_test.py
```diff
@@ -0,0 +1,115 @@
...
+            DEFAULT_NOTIFICATION_FETCH_COUNT,
+            0,
+            mock.ANY, # Unfortunately time flies away in between the call.
+            False
+        )
+        # With all optional paramters.
+        url = f'notification/{user_id}/fetch?project=listenbrainz,metabrainz&count=3&offset=1&until_ts={until_ts.timestamp()}&unread_only=t'
+        response = self.client.get(url)
+        self.assert200(response)
+        mock_fetch.assert_called_with(
```
fettuccinae, line 30: For the endpoint tests, I've only checked how the call is made to the db function.
Do i need to verify the return value too?

metabrainz/db/notification.py
```diff
@@ -53,10 +53,10 @@ def fetch_notifications(user_id: int, projects: Optional[Tuple[str, ...]]=None,
                         AND (:unread_only = FALSE OR read = FALSE)
                 ORDER BY
                         created DESC
-                LIMIT
-                        :count
                 OFFSET
                         :offset
+                LIMIT
```
fettuccinae, line 58: I think `OFFSET` should be set before `LIMIT` because if we do it the other way( setting LIMIT before OFFSET), we'll be skipping rows after limiting the result.

metabrainz/notifications/views_test.py
mayhem: yes, please

metabrainz/db/notification.py
mayhem: The order is required by SQL, you can't have offset after limit. :)

fettuccinae: Ohh, I think ive seen it done the other way in db/payments.py . 
Should i make another pr replacing depricated methods (utc.now, logger.warn) and changing that query?

mayhem: Which query is that? Got link? But cleanup PRs are most welcome!

fettuccinae: [metabrainz/db/payments.py](https://github.com/metabrainz/metabrainz.org/blob/19acb9456d20ab873afdd08aa2595c9991f7d3cb/metabrainz/db/payment.py#L40)
[And a few in LB too](https://github.com/search?q=repo%3Ametabrainz%2Flistenbrainz-server+%2F%28%3F-i%29OFFSET%2F+language%3APython+&type=code)

mayhem (approved):
Looks good!

PR #515, opened 2025-06-18
Remove fixed height properties in tier styles

amCap1712:
I think this broke because MeB.org's default font size was increased some months back.

Before:
image

After:
image


mayhem (approved):
Thanks!!

MonkeyDo (approved):
:rocket:

PR #517, opened 2025-07-02
Clean-up PR

fettuccinae:
Replaced depricated methods

1. datetime.utcnow() with datetime.now(timezone.utc)
  https://docs.python.org/3/library/datetime.html#datetime.datetime.now

2. logger.warn() with logger.warning()
  https://docs.python.org/3.11/library/logging.html#logging.Logger.warning


mayhem (approved):
thanks!

PR #519, opened 2025-07-09
Send notifications immediately to users with digest disabled

fettuccinae:
Refactored code to send non important notifications.
Added a function that returns notifications which are to be sent to users with digest disabled.


fettuccinae:
@mayhem

mayhem (approved):
Looks good!

PR #521, opened 2025-07-21
Add a function to mark notifications as "read"

fettuccinae:
Changed insert_notification function to use execute_values(psycopg2) instead of execute_many(sqlalchemy) because execute_many doesnt support "RETURNING", and we need the IDs of the notifications that were inserted so that we can mark them as read later.

Added a function to mark notifications as read (used after sending notifications).

Fixed hset bug by converting dictionary object into a JSON string.

Changed and added tests to accommodate these changes


fettuccinae:
@mayhem

mayhem:
When you PR is ready to review, please request a review from me, otherwise I won't know if the PR is ready!

metabrainz/db/notification.py
```diff
@@ -157,8 +169,9 @@ def _prepare_notifications(notifications: List[dict]) -> List[dict]:
                 "project": notif["project"],
                 "important": notif["important"],
                 "expire_age": notif["expire_age"],
+                # If we use pgsql's default UUID column, test cases fail due to "uuid-ossp" extension not being found.
```
mayhem, line 172: This is installed on all of our servers, so no worry there.

mayhem (approved):

mayhem:
image
 You don't have this?

fettuccinae:
> You don't have this?

Nope, Not on PR's which have 0 reviews. I can request a re-review tho.

PR #529, opened 2025-10-01
Add MTCaptcha to third-parties in privacy policy

yvanzo:
MTCaptcha does replace reCAPTCHA for MusicBrainz account registration.

Reference: https://tickets.metabrainz.org/browse/MBS-10039


yvanzo:
Note that reCAPTCHA seems to be in use for MetaBrainz account registration still.
See https://github.com/search?q=org%3Ametabrainz%20recaptcha&type=code

It has to be either entirely removed or deact

mwiencek (approved):

PR #590, opened 2026-05-11
consul_config: fall back to pgbouncer-master when pgbouncer-slave is …

zas:
…unavailable

When pgbouncer-slave is down, MB_DATABASE_URI was set to an empty string, causing the OAuth app to crash on startup. Fall back to pgbouncer-master instead.


MonkeyDo (approved):
Looks good.
Not sure why it wasn't set up like that before, but good that it's fixed

mwiencek (approved):

PR #593, opened 2026-06-10
Add internationalization (i18n) support

abhinavohri:
## Overview
Implements localization for the site so pages can be served in different languages. 

## Methodology
The language is resolved from a cookie on the server and used directly by Flask-Babel to translate server-rendered content. For the React side, the locale is injected into the page as a global prop, which the frontend uses to initialize i18next when the app boots — so both sides translate off the same locale.

## Pending
- Generate translated catalogs


amCap1712 (approved):

MonkeyDo (commented):
Sorry for the late review!

Looks great, thanks for working on this Abhinav

PR #601, opened 2026-07-16
Make more stuff translatable

reosarevok:
Please double check if this is the right way to do these things - I've aped what I saw @MonkeyDo do for the combined strings. I checked locally and they seem to work, anyway. I also made some big policy docs translatable, moved the URLs into variables but not quite sure what to do if anything with the email addresses.

### DIFF
--- a/metabrainz/templates/index/gdpr.html
+++ b/metabrainz/templates/index/gdpr.html
@@ -7,200 +7,209 @@
 
   <h2>{{ _('Introduction') }}</h2>
   <p>
-    The General Data Protection Regulation is a complex EU regulation that stipulates many points for protecting private data of
+    {{ _('The General Data Protection Regulation is a complex EU regulation that stipulates many points for protecting private data of
     users on the Internet. Even though this is an EU regulation, it has a worldwide impact due to the nature of the Internet.
     The MetaBrainz Foundation with its collection of projects is also affected by this regulation. We’ve been learning and adapting
     our sites to be compliant with the GDPR – sadly this regulation isn’t entirely black and white and there is an incredible
-    amount of room left for interpretation of these rules.
+    amount of room left for interpretation of these rules.') }}
   </p>
   <p>
-    The good news is that this regulation is roughly in line with our established practices: We’ve always held private information
+    {{ _('The good news is that this regulation is roughly in line with our established practices: We’ve always held private information
     in a high regard and applied the sort of rules to ourselves as we wish to have our own private data treated. Luckily, this
     makes our compliance effort considerably easier. We’ve made two significant changes to how we treat your data and also adopted
     terminology as used in the GDPR in order to use the same languages that many other sites are now adopting. Please keep reading
-    to find out the exact details of what we are doing to comply.
+    to find out the exact details of what we are doing to comply.') }}
   </p>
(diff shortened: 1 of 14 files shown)


metabrainz/templates/index/privacy.html
```diff
@@ -9,137 +9,139 @@ <h2>{{ _('Summary') }}</h2>
...
   <p>
-    <i>The above information assumes that you are using "normal" web browser settings, whereby images are always loaded and HTTP referrer 
-    information is always sent.</i>
+    {{ _('<i>The above information assumes that you are using "normal" web browser settings, whereby images are always loaded and HTTP referrer 
+    information is always sent.</i>') }}
   </p>
 
   <p>
-    It is impractical to constantly modify this privacy policy when MusicBrainz changes which third party content it utilizes. We will 
+    {{ _('It is impractical to constantly modify this privacy policy when MusicBrainz changes which third party content it utilizes. We will
```
MonkeyDo:
```suggestion
    {{ _('It is impractical to constantly modify this privacy policy when MetaBrainz projects change which third party content they utilize. We will 
```

metabrainz/templates/index/gdpr.html
```diff
@@ -7,200 +7,209 @@ <h1 class="page-title">{{ _('GDPR Compliance Statement') }}</h1>
...
+     <li>{{ _('Implement sitewide user account deletion in
          <a href="https://tickets.metabrainz.org/browse/MBS-9680">MusicBrainz</a>,
          <a href="https://tickets.metabrainz.org/browse/AB-337">AcousticBrainz</a>,
          <a href="https://tickets.metabrainz.org/browse/BB-260">BookBrainz</a>,
          <a href="https://tickets.metabrainz.org/browse/MEB-103">the MetaBrainz Site</a>,
          <a href="https://tickets.metabrainz.org/browse/LB-354">ListenBrainz</a>, and
-         <a href="https://tickets.metabrainz.org/browse/CB-307">CritiqueBrainz</a>.
+         <a href="https://tickets.metabrainz.org/browse/CB-307">CritiqueBrainz</a>.') }}
      </li>
   </ul>
```
MonkeyDo, line 78: We need to review this section. Account deletion is now on the MeB side, and the other projects have it implemented through webhooks (at least LB, MB, BB and CB)

Might even be worth to remove all mentions of account deletion. Maybe keep the marking-a-donation-as-private part, but there is already a mention of that at lines [168-170](https://github.com/metabrainz/metabrainz.org/pull/601/changes#diff-35343d3666540106873bbb4fef60a90723c79135cfe2990fac504b73238c4868R168-R170)

MonkeyDo (changes requested):
One small change, and one big change that requires review after deploying the new OAuth system

metabrainz/templates/index/gdpr.html
reosarevok: Yes, I think we should just say that deleting your account in MeB will delete all accounts in our other MetaBrainz projects, *but* we need to figure out what, if anything, happens to Weblate / Jira / Discourse / any other places that support MeB accounts.

MonkeyDo (approved):
Fanks!

PR #607, opened 2026-07-27
Fixed token revocation URL

amCap1712:


MonkeyDo (approved):
OMG, so simple!
I had not seen the discrepancy

PR #609, opened 2026-07-27
Fix webhook delivery queue state boundaries

amCap1712:
Persist webhook delivery state consistently across the database and Celery publication boundary so broker or worker failures cannot leave deliveries permanently pending.

- centralize task publication for new, scheduled, and administrator-triggered deliveries
- return broker publication failures to a retryable failed state without consuming an HTTP retry
- recover stale pending rows and processing rows abandoned by dead workers
- require the related webhook to remain administrator-enabled for every scheduled retry state
- atomically claim pending deliveries before sending to suppress duplicate HTTP requests
- compare status and update timestamp when requeueing so stale callers cannot resurrect claimed or completed deliveries
- release processing claims before retrying task-level failures
- report concurrent retry transitions accurately in the administration interface

### DIFF
--- a/metabrainz/admin/webhooks.py
+++ b/metabrainz/admin/webhooks.py
@@ -14,6 +14,7 @@ from metabrainz.admin.forms import RetryDeliveryForm
 from metabrainz.model import db
 from metabrainz.model.webhook import Webhook, WebhookDelivery, EVENT_USER_CREATED, EVENT_USER_DELETED, \
     EVENT_USER_UPDATED
+from metabrainz.webhooks.tasks import enqueue_webhook_delivery
 
 
 def validate_webhook_url(form, field):
@@ -201,20 +202,31 @@ class WebhookDeliveryModelView(AdminModelView):
             
             count = 0
             skipped = 0
+            failed = 0
             for delivery in query.all():
                 if delivery.status in ["failed", "pending"]:
-                    delivery.status = "pending"
-                    delivery.error_message = None
-                    count += 1
+                    try:
+                        if enqueue_webhook_delivery(delivery):
+                            count += 1
+                        else:
+                            skipped += 1
+                    except Exception as error:
+                        failed += 1
+                        current_app.logger.error(
+                            "Failed to enqueue webhook delivery %s: %s",
+                            delivery.id,
+                            error,
+                            exc_info=True,
+                        )
                 else:
                     skipped += 1
-            
-            db.session.commit()
-            
+
             if count > 0:
                 flash(f"Queued {count} delivery(ies) for retry.", "success")
             if skipped > 0:
                 flash(f"Skipped {skipped} delivery(ies) (only failed/pending can be retried).", "info")
+            if failed > 0:
+                flash(f"Failed to queue {failed} delivery(ies). They will be retried automatically.", "error")
         except Exception as e:
             db.session.rollback()
             current_app.logger.error(f"Error retrying deliveries: {str(e)}", exc_info=True)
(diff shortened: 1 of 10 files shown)


metabrainz/admin/webhooks.py
```diff
@@ -201,20 +202,31 @@ def action_retry_failed(self, ids):
...
-            
-            db.session.commit()
-            
+
             if count > 0:
                 flash(f"Queued {count} delivery(ies) for retry.", "success")
             if skipped > 0:
                 flash(f"Skipped {skipped} delivery(ies) (only failed/pending can be retried).", "info")
+            if failed > 0:
+                flash(f"Failed to queue {failed} delivery(ies). They will be retried automatically.", "error")
```
reosarevok, line 229: Is there a limit on how many times these are retried? If not maybe we should prioritize `pending` over `failed` so that new ones that do not error still go through fine if there's an error leading to more and more `failed` ones accumulating?

amCap1712: 5 max retries.

metabrainz/admin/webhooks.py
```diff
@@ -201,20 +202,31 @@ def action_retry_failed(self, ids):
             
             count = 0
             skipped = 0
+            failed = 0
             for delivery in query.all():
                 if delivery.status in ["failed", "pending"]:
```
MonkeyDo, line 207: The method is called retry_failed, but we also process pending ones?

EDIT: Never mind, I saw the tests for this below, makes sense

metabrainz/webhooks/tasks.py
```diff
@@ -34,31 +56,132 @@ def deliver_webhook(self, delivery_id: str):
...
+
+
+def _enqueue_webhook_delivery_if(
+    delivery_id,
+    *eligibility_conditions,
+) -> bool:
+    """
+    Atomically make an eligible delivery pending, then publish its task.
+
+    Duplicate queued tasks are safe because delivery workers atomically claim
```
MonkeyDo, line 152: I saw some of the failed deliveries are due to LB responmding with 500 because of re-processing an event, it looks like.

I don't know if this is because you tried re-sending webhooks at some point that had been processed, or if it is due to an issue with the task processing, but thought I would flag it in case it is related

```
Webhook received: event=user.created, delivery=d774b0e1-e17c-4b33-ad43-dd290d159015, attempt=3, user_agent=MetaBrainz-Webhooks/1.0
Error processing webhook: event=user.created, delivery=d774b0e1-e17c-4b33-ad43-dd290d159015:
[....]
(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "user_id_ndx_listen_user_metadata_new"
DETAIL:  Key (user_id)=(149261) already exists.

[SQL: INSERT INTO listen_user_metadata VALUES (%(user_id)s, 0, NULL, NULL, NOW())]
[parameters: {'user_id': 149261}]
```

metabrainz/webhooks/tasks.py
```diff
@@ -34,31 +56,132 @@ def deliver_webhook(self, delivery_id: str):
...
+
+def _publish_webhook_delivery(delivery_id, enqueued_at: datetime) -> None:
+    """Publish a delivery task and recover safely if broker publication fails."""
+    try:
+        deliver_webhook.apply_async(
+            args=[str(delivery_id)],
+            queue="webhooks",
+        )
+    except Exception as error:
+        db.session.rollback()
```
MonkeyDo, line 115: What does this roll back?

MonkeyDo (commented):
Well, this has become quite complex.
Not sure there's much we can do about it at this point, but in any case, the improvements in this PR do improve the processing and retry mechanisms and other failure modes 👍

metabrainz/webhooks/tasks.py
amCap1712: Yes, clients need to be made idempotent.

metabrainz/webhooks/tasks.py
amCap1712: Any stuck state.

PR #610, opened 2026-07-28
Simplify webhook circuit breaking with Redis failure windows

amCap1712:
Replace the process-local circuit breaker state machine with a Redis sorted set keyed by webhook ID. Store each endpoint failure using its timestamp as the score, count only failures inside the configured rolling window, prune expired entries on writes, and expire inactive keys automatically.

Defer deliveries while the failure threshold is exceeded without consuming their HTTP retry budget, and schedule the next attempt for when enough failures have aged out. Distinguish remote endpoint errors from internal delivery errors so signing or persistence failures do not disable otherwise healthy endpoints, while both paths retain bounded delivery retries.


metabrainz/webhooks/circuit_breaker.py
```diff
@@ -1,127 +1,82 @@
...
-            if self._state == CircuitBreakerState.HALF_OPEN:
-                # Success in half-open state means we can close the circuit
-                self._state = CircuitBreakerState.CLOSED
+        self.failure_window = failure_window
+
+    def _recent_failures(self) -> list:
+        now = time.time()
+        # Scores are timestamps, so this returns only failures still inside the
+        # rolling window. Older entries do not affect the circuit state.
+        return self.redis_client.zrangebyscore(
```
mwiencek: Is that using https://redis.io/docs/latest/commands/zrangebyscore/ directly? It says it's been deprecated, unless the Python API is already doing "ZRANGE with the BYSCORE argument" underneath.

mwiencek (commented):
A helpful thing would be some kind of design document for the webhooks that concisely explains what the intended behavior for the retries / circuit breaking is. It's kind of impenetrable from the code alone. I get the idea of moving things from in-process state to Redis at least. And there are tests, so I trust you to deploy this if you're confident in it.

metabrainz/webhooks/circuit_breaker.py
amCap1712: Fixed, thanks!

mwiencek (approved):
:shipit: if this will resolve our current production issues, the documentation was a suggestion for later

PR #611, opened 2026-07-28
Improve admin search and supporter signup validation

amCap1712:
Make user and supporter searches include pending email addresses, trim search input, escape wildcard characters, retain partial matching, and rank case-insensitive exact matches before broader results.

Enable search on the old-username admin view and include the supporter's username and pending-or-confirmed contact email in commercial signup notifications.

Keep the commercial supporter website field optional while validating supplied values as HTTP(S) URLs in both the backend form and signup UI.

### DIFF
--- a/frontend/js/src/forms/SignupCommercial.tsx
+++ b/frontend/js/src/forms/SignupCommercial.tsx
@@ -115,9 +115,12 @@ function SignupCommercial({
     org_desc: Yup.string().required(
       t("You need to provide description of your organization.")
     ),
-    website_url: Yup.string().required(
-      t("You need to specify website of the organization.")
-    ),
+    website_url: Yup.string()
+      .url(t("Website URL must be a valid URL."))
+      .matches(/^https?:\/\//i, {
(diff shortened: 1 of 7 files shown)


frontend/js/src/forms/SignupCommercial.tsx
```diff
@@ -115,9 +115,12 @@ function SignupCommercial({
     org_desc: Yup.string().required(
       t("You need to provide description of your organization.")
     ),
-    website_url: Yup.string().required(
-      t("You need to specify website of the organization.")
-    ),
+    website_url: Yup.string()
+      .url(t("Website URL must be a valid URL."))
+      .matches(/^https?:\/\//i, {
+        message: t("Website URL must be a valid URL."),
```
reosarevok, line 121: Why is the error string here twice?

amCap1712: Two validators, one is to check its a URL. second is to check its starts with https://, each validator gets its own message.

PR #612, opened 2026-07-28
Fix OAuth token cleanup performance

amCap1712:
Replace the monolithic OAuth token DELETE statements with bounded, lock-aware batches. Select up to 5,000 eligible token IDs at a time, ordered by issued_at and id, lock them with FOR UPDATE SKIP LOCKED, delete them, and commit each batch independently.

Add partial indexes on (issued_at, id) for both OAuth token tables so the database can efficiently locate and order cleanup candidates. Build the indexes concurrently when upgrading existing installations and add them to the bootstrap DDL for newly initialized databases.

### DIFF
--- a/metabrainz/oauth/tasks.py
+++ b/metabrainz/oauth/tasks.py
@@ -3,51 +3,91 @@ from typing import Any
 
 from celery import shared_task
 from flask import current_app
-from sqlalchemy import text
+from sqlalchemy import Table, and_, delete, or_, select
 
-from metabrainz.model import db
+from metabrainz.model import OAuth2AccessToken, OAuth2RefreshToken, db
 
 
-DELETE_OLD_TOKENS_SQL = """
-DELETE FROM oauth.{table_name}
-WHERE
-    (
-        COALESCE(revoked, FALSE) = TRUE
-        AND issued_at IS NOT NULL
-        AND issued_at < :cutoff_date
+TOKEN_CLEANUP_BATCH_SIZE = 5_000
+
+
+def _cleanup_token_table(
+    token_table: Table,
+    cutoff_date: datetime,
+    batch_size: int,
+) -> int:
+    deleted_total = 0
+    expires_at = (
+        token_table.c.issued_at
+        + token_table.c.expires_in * timedelta(seconds=1)
     )
-    OR (
-        issued_at IS NOT NULL
-        AND expires_in IS NOT NULL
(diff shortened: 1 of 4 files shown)


metabrainz/oauth/tasks.py
```diff
@@ -3,51 +3,91 @@
...
+        .limit(batch_size)
+        .with_for_update(skip_locked=True)
+        .cte("candidates")
     )
-    OR (
-        issued_at IS NOT NULL
-        AND expires_in IS NOT NULL
-        AND issued_at + (expires_in * INTERVAL '1 second') < :cutoff_date
+    statement = delete(token_table).where(
+        token_table.c.id.in_(select(candidates.c.id))
```
MonkeyDo, line 43: Yeah, this is more or less like the example that bitmap shared [from the PG docs](https://www.postgresql.org/docs/18/sql-delete.html#id-1.9.3.100.9):

```psql
WITH delete_batch AS (
  SELECT l.ctid FROM user_logs AS l
    WHERE l.status = 'archived'
    ORDER BY l.creation_date
    FOR UPDATE
    LIMIT 10000
)
DELETE FROM user_logs AS dl
  USING delete_batch AS del
  WHERE dl.ctid = del.ctid;
```

metabrainz/oauth/tasks.py
```diff
@@ -3,51 +3,91 @@
...
     )
-"""
+
+    while True:
```
MonkeyDo, line 46: Can this hang in a loop in any unforeseen scenario?

MonkeyDo (approved):
Yes, that'll do it

mwiencek:
I'm trying to understand the behavior here with respect to refresh tokens. Those all have expires_in = 3600, but IIRC you mentioned before that the intent is that they never expire, at least currently. This would seem to just delete them after 7 days (unless I've misunderstood the code). Is that intended?

mwiencek (commented):
I ran the celery task against some test data and it does seem to work as expected. I have `log_statement = all` in my PG config so I was able to check the actual queries it's doing and they do seem to use the added indexes. So that's all good. I'm still wondering how refresh tokens are supposed to work, and if it's consistent with what's documented in https://github.com/metabrainz/metabrainz.org/blob/master/docs/oauth/authorization-code-grant.rst

amCap1712:
Will fix refresh tokens in subsequent PR.

PR #613, opened 2026-07-29
Misc fixes after accounts migration

amCap1712:
- Fix OAuth refresh token expiration configuration
- Validate account links and bcrypt password length
- Include standard context in application logs
- Send supporter tier notification after commit
- Notify supporters at their available email address even if unconfirmed
- Use redis/service from consul to separate prod/test redis

### DIFF
--- a/frontend/js/src/forms/utils.tsx
+++ b/frontend/js/src/forms/utils.tsx
@@ -1,6 +1,10 @@
 import React, { JSX } from "react";
 import { Field, FieldConfig, useField } from "formik";
 import { useTranslation } from "react-i18next";
+import {
+  passwordExceedsBcryptByteLimit,
+  setPasswordByteLimitValidity,
+} from "./password";
 
 export type FormLevelAlertProps = {
   errors: Record<string, string> | null | undefined;
@@ -20,18 +24,31 @@ export function getAuthPageUrl(path: "/login" | "/signup"): string {
 export type TextInputProps = JSX.IntrinsicElements["input"] &
   FieldConfig & {
     label: string;
+    inputRef?: React.RefObject<HTMLInputElement | null>;
   };
 
-export function TextInput({ label, children, ...props }: TextInputProps) {
+export function TextInput({
+  label,
+  children,
+  inputRef,
+  ...props
+}: TextInputProps) {
   const [field, meta] = useField(props);
   const hasError = meta.touched && meta.error;
+  const inputProps = { ...props };
+  delete inputProps.validate;
   return (
     <div className={`form-group ${hasError ? "has-error" : ""}`}>
       <label className="col-sm-4 control-label" htmlFor={props.id}>
         {label} {props.required && <span style={{ color: "red" }}>*</span>}
       </label>
       <div className="col-sm-5">
-        <input className="form-control" {...field} {...props} />
+        <input
+          className="form-control"
+          {...field}
+          {...inputProps}
+          ref={inputRef}
+        />
       </div>
       {children}
       {hasError ? (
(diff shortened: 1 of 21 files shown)


metabrainz/templates/users/reauthenticate.html
```diff
@@ -26,3 +31,21 @@ <h1 class="page-title">{{ _('Confirm Password') }}</h1>
...
+    document.addEventListener("DOMContentLoaded", function () {
+      var passwordInput = document.querySelector("[data-max-utf8-bytes]");
+      var byteLimitError = {{ _("Password must not exceed 72 bytes.")|tojson }};
+
+      function updatePasswordValidity() {
+        var byteLength = new TextEncoder().encode(passwordInput.value).byteLength;
+        passwordInput.setCustomValidity(byteLength > 72 ? byteLimitError : "");
+      }
+
+      passwordInput.addEventListener("input", updatePasswordValidity);
```
MonkeyDo: Might be worth throttling this for a text input, rather than running the function for each stroke.

A simple throttling function [from stackoverflow](https://stackoverflow.com/a/27078401/4904467):
```js
function throttle (callback, limit) {
    var waiting = false;
    return function () {
        if (!waiting) {
            callback.apply(this, arguments);
            waiting = true;
            setTimeout(function () {
                waiting = false;
            }, limit);
        }
    }
}
```
then:
```js
var throttledUpdatePasswordValidity = throttle(updatePasswordValidity, 500);
passwordInput.addEventListener("input", throttledUpdatePasswordValidity);
```

metabrainz/supporter/views_test.py
```diff
@@ -377,18 +386,26 @@ def test_commercial_signup_notification_includes_username(self):
...
-                data_usage_desc="Testing notifications",
-                org_name="Notification Organization",
-                user=user,
-            )
+            send_supporter_signup_notification(supporter)
 
         notification_text = send_mail.call_args.kwargs["text"]
         self.assertIn("Username: notification-user", notification_text)
         self.assertIn("Contact email: notification@example.com", notification_text)
+        self.assertIn(
```
MonkeyDo, line 405: The test should be renamed as it tests more than the username

frontend/js/src/forms/utils.tsx
```diff
@@ -195,9 +232,57 @@ export function AuthCardPasswordInput({ ...props }: AuthCardTextInputProps) {
...
+        }
+        return validate?.(value);
+      }}
+      onInput={(event) => {
+        updateByteLimitValidity(event.currentTarget);
+        onInput?.(event);
+      }}
     />
   );
 }
```
MonkeyDo: I suggest using the AuthCardPasswordInput instead everywhere.
It has a good user feature (show/hide password) and the same byte limit validation anyway, so why not use that directly?
As it stands, AFAICT this new component is basically a copy but without the show/hide button, which we could make parametric if we wanted to remove it (I don't suggest we do)

frontend/js/src/forms/utils.tsx
```diff
@@ -20,18 +24,31 @@ export function getAuthPageUrl(path: "/login" | "/signup"): string {
...
   FieldConfig & {
     label: string;
+    inputRef?: React.RefObject<HTMLInputElement | null>;
   };
 
-export function TextInput({ label, children, ...props }: TextInputProps) {
+export function TextInput({
+  label,
+  children,
+  inputRef,
```
MonkeyDo: Don't you need like a forwardRef for this to work?

EDIT: Never mind, React 19 has [deprecated forwardRef](https://es.react.dev/reference/react/forwardRef) and we can just pass a ref prop directly. Good to know!

MonkeyDo (commented):
Thanks for working on these improvements.

We should p[ut this on test.meb and test it manually, ensuring email are sent, supporters without emails cna be rejected, etc.
Let me know when it is up on test, I'd be happy to help test the supporter signup flows

PR #618, opened 2026-08-04
Reject non-UUID session identifiers in load_user

amCap1712:
User.get_id() returns login_id, a uuid column, but session and remember cookies issued before that change still carry the integer primary key. flask-login passes the stored value straight to the user loader, so those cookies reached the database as "user".login_id = '560'::uuid and raised sqlalchemy.exc.DataError, turning every request from an affected browser into a 500.

Parse the identifier as a UUID before querying and return None when it does not parse. flask-login then treats the request as anonymous, so a stale cookie logs the user out once instead of erroring. REMEMBER_COOKIE_DURATION is a year, so these cookies keep arriving for a long time and the check needs to stay.

### DIFF
--- a/metabrainz/user/__init__.py
+++ b/metabrainz/user/__init__.py
@@ -1,4 +1,5 @@
 from functools import wraps
+from uuid import UUID
 
 from flask import abort, flash, redirect, request, url_for
 from flask_login import LoginManager, current_user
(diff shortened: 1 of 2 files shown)


metabrainz/user/__init__.py
```diff
@@ -15,10 +16,14 @@
...
 @login_manager.user_loader
 def load_user(login_id):
+    # Sessions created before get_id() switched from the integer primary key to login_id
+    # still carry an integer here. Reject anything that isn't a UUID before it reaches the
+    # database, otherwise postgres raises a DataError on the ::uuid cast.
     try:
-        return User.get(login_id=login_id)
-    except (ValueError, TypeError):
+        login_id = str(UUID(str(login_id)))
+    except (ValueError, TypeError, AttributeError):
```
mwiencek, line 24: Can that throw `AttributeError` as written? I'm curious how.

mwiencek (approved):

metabrainz/user/__init__.py
amCap1712: Hmm, actually yeah it wouldn't happen here. will remove.

PR #620, opened 2026-08-11
Restrict certain scopes to certain OAuth clients

amCap1712:
A scope can now be marked restricted, and only the clients explicitly granted it may request it; every other client is refused with invalid_scope.

oauth.scope gains a restricted flag and oauth.l_client_scope records which clients hold a grant for which restricted scope. Nothing changes for an unrestricted scope never consults the grant table.

Enforcement sits in two places because the grants do not all validate scope the same way. RestrictedScopeMixin hooks validate_requested_scope() on the authorization code, implicit, OpenID implicit and client credentials grants and raises invalid_scope naming the refused scopes, which is where a client developer gets a message they can act on. authlib sets request.client before that call in each of those grants, so the check can see who is asking.

The refresh token grant never revalidates the requested scope - it only checks the request against the scope of the token being refreshed - so a mixin alone would let a client keep minting tokens for a restricted scope forever after its grant is withdrawn. OAuth2Client.get_allowed_scope therefore returns None, which authlib turns into invalid_scope, and every grant reaches it through token generation. A client can still narrow the request itself with a scope parameter; otherwise the user has to authorize the application again.

Restricted scopes are left out of scopes_supported in the discovery document.


mwiencek (approved):
LGTM. The added tests look pretty thorough. I also tested the admin UI + the client credentials grant locally, and it seems to work as expected based on the restricted scopes.

docs/oauth/scopes.rst
```diff
@@ -77,6 +77,10 @@ ListenBrainz
    * - ``listenbrainz:submit-listens``
      - Submit listens to ListenBrainz.
      - *Submit listens to ListenBrainz.*
+   * - ``listenbrainz:connect-services``
```
MonkeyDo: What did "Restricted scopes are left out of scopes_supported in the discovery document." mean?
I misunderstood it as "not-documented" but that's clearly not the case.

metabrainz/oauth/implicit_grant.py
```diff
@@ -3,8 +3,10 @@
 from authlib.oauth2.rfc6749 import grants, InvalidRequestError
 from flask import render_template
 
+from metabrainz.oauth.restricted_scope import RestrictedScopeMixin
 
-class ImplicitGrant(grants.ImplicitGrant):
+
+class ImplicitGrant(RestrictedScopeMixin, grants.ImplicitGrant):
```
MonkeyDo, line 9: Since this mixin is repeated for every type of grant, could we have a base class that uses it, and extend it for all others?
Not sure it matters, but I wonder if implementing a new grant it could be easy to forget to add the mixin. That being said it could also be easy to forget to extend the base class, so probably not that important, especially since we're not adding a new grant type every week

MonkeyDo (approved):
Sorry I'm late to the party, I reviewed this over two days. Took me a while as I needed to look some stuff up to understand everything.
Looks good.

One small comment/question that can be ignored.

PR #623, opened 2026-08-21
Fix mail delivery and harden payment webhooks                                                                                                                                                                         

amCap1712:
- copy the email delivery helper from brainzutils into MetaBrainz
  - use bare recipient addresses for SMTP headers and envelopes
  - add an SMTP timeout and consistently wrap delivery errors
  - support pending email addresses in username recovery
  - prevent email failures from breaking signup, verification, profile, supporter, and payment flows
  - preserve committed payments when receipt delivery fails
  - normalize lowercase Stripe boolean metadata before storing payments
  - display profile email conflicts instead of silently redirecting

### DIFF
--- a/metabrainz/user/email.py
+++ b/metabrainz/user/email.py
@@ -3,7 +3,7 @@ from datetime import datetime
 
 from flask import url_for, request, render_template, current_app
 
-from brainzutils.mail import send_mail
+from metabrainz.mail import send_mail
 from metabrainz.model.user import User
 
 VERIFY_EMAIL = "verify-email"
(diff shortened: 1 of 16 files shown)


metabrainz/user/email.py
```diff
@@ -18,9 +18,11 @@ def create_email_link_checksum(purpose: str, user_id: int, email: str, timestamp
     return m.hexdigest()
 
 
-def _send_user_email(username: str, email: str, subject: str, content: str):
+def _send_user_email(email: str, subject: str, content: str):
     if not current_app.config["DEBUG"]:
```
mwiencek, line 22: (took me a while to figure out I had to disable `DEBUG` mode to use my local SMTP server, is this restriction necessary? :))

mwiencek (approved):
I've only tested the email changes locally. Was able to reproduce the verification failures with the test cases you added, on master, and they work fine on this branch. Haven't checked the Stripe changes, so up to you if you're confident to merge those.

metabrainz/user/email.py
amCap1712: It leads to errors in development if you don't have SMTP server setup but I guess we could check for that instead of DEBUG then. Will fix separately.

PR #626, opened 2026-08-28
MEB-190: Localize OAuth authorization pages (ui_locales + error messages)

zas:
## Problem

The OAuth2/OIDC authorization and error pages are rendered only in the language
selected via the website's `lang` cookie (`get_locale()`, wired through
`Babel(app, locale_selector=get_locale)`). OAuth clients such as MusicBrainz
Picard open the authorization URL in an **external browser** where no `lang`
cookie exists, so a user whose client is configured for another language (e.g.
French) always sees the authorization and error pages in English.

## Changes

**1. Honor the OpenID Connect `ui_locales` request parameter**

`get_locale()` now falls back to `ui_locales` (a space-separated,
preference-ordered list of BCP 47 tags, per OIDC Core 1.0 §3.1.2.1) when no
supported `lang` cookie is present, matching on the primary language subtag
against the supported locales (`en`, `es`, `fr`, `de`). An explicit cookie still
takes precedence, and unsupported/unknown locales fall back to the default
(never an error, as the spec requires).

**2. Localize OAuth authorization error messages by error code**

OAuth error pages previously showed only the untranslated English
`error_description`. Per RFC 6749 §4.1.2.1, `error_description` is a
developer-facing, ASCII-only field and must not carry localized text, so the
user-facing message is instead derived from the stable OAuth/OIDC error *code*
(`access_denied`, `invalid_scope`, etc.) via gettext. The error handler passes a
translated `message` prop chosen by code (None for unknown codes); the
`OAuthError` React component shows it prominently and keeps the raw
`code: description` as secondary technical detail.

## Testing

- New unit tests: `metabrainz/i18n_test.py` (12 cases: `match_ui_locales` +
  `get_locale` precedence) and `metabrainz/errors_test.py` (3 cases: error-code
  message map). All pass via `./test.sh`.
- Frontend `npm run type-check` and `eslint` clean on the changed component.

## Notes

- New gettext strings will be picked up by the normal translation workflow
  (Weblate / `messages.pot` regeneration) — not included here.
- Scope descriptions on the consent page are still English; that is a separate,
  larger change (DB-stored descriptions) tracked in its own ticket.

Ticket: MEB-190

### DIFF
--- a/docs/oauth/authorization-code-grant.rst
+++ b/docs/oauth/authorization-code-grant.rst
@@ -60,6 +60,15 @@ Redirect the user's browser to the authorization endpoint.
       ``auto``, if the user has already granted the requested scopes to your
       client, consent is skipped and a code is returned immediately. With
       ``force`` the consent screen is always shown.
+   :query ui_locales: Optional. Space-separated list of preferred BCP 47
+      language tags, in order of preference, as defined by `OpenID Connect
+      Core 1.0 section 3.1.2.1
+      <https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest>`_,
+      e.g. ``de-AT de en``. The first tag matching a language MetaBrainz
+      supports is used for the sign in, sign up, consent and error pages of
+      this flow. A language the user has explicitly chosen on the site takes
+      precedence. The supported values are listed as ``ui_locales_supported``
+      in the :doc:`discovery document <openid-connect>`.
 
 Example:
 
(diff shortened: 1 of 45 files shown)


MonkeyDo (commented):
Did not test (and not sure how to test this with picard, so I'll leave that to you), but reviewed carefully and it looks all fine.

amCap1712 (approved):

PR #627, opened 2026-08-28
Fix support page tier cards overlapping due to unequal heights

anshg1214:
# Problem

## What is the issue?
On the Support Us page, the Silver tier card wraps underneath the Bronze card instead of starting on a new row.

image

## Why does it happen?
The tier cards use Bootstrap's float-based grid. When cards in a row have unequal heights, the float clearing breaks. A tall card prevents the next card from clearing past it, causing it to snag and wrap into the wrong position.

* [x] I have reviewed the code and tested it locally

# AI usage

* [ ] I did not use any AI
* [x] I have used AI in this PR (add more details below)

#### If you did use AI:
* [ ] I used AI tools for communication
* [x] I used AI tools for coding
* [x] I understand all the changes made in this PR

### DIFF
--- a/frontend/css/main.less
+++ b/frontend/css/main.less
@@ -125,8 +125,24 @@ img.pull-right {
 #tiers {
   margin-top: 40px;
   #primary-tiers {
+    .row {
+      display: flex;
+      flex-wrap: wrap;
+    }
     .tier {
+      display: flex;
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Thanks!

